PR #1, opened 2021-03-24
feat: scaffolded code

Tulun:
-Ran CRA w/ TypeScript template.
-Added in basic folder structure, router, and css lib (tailwind).


src/reportWebVitals.ts
```diff
@@ -0,0 +1,15 @@
+import { ReportHandler } from 'web-vitals';
+
+const reportWebVitals = (onPerfEntry?: ReportHandler) => {
```
daywiss, line 3: not familiar with this, what is it?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
+    "noFallthroughCasesInSwitch": true,
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react-jsx"
```
daywiss, line 18: @Gamaranto how does this look?

also do we want to add additional linting / prettifying?

crago.config.ts
```diff
@@ -0,0 +1,8 @@
+// craco.config.js
+module.exports = {
```
daywiss, line 2: is this file craco or crago? also not familiar with this

daywiss (commented):
looks good to me. @Gamaranto just want you to see if theres anything you would like to pull in based on the claim dapp

crago.config.ts
Tulun: Seems like it was a bridge with a breaking React 17 change. removed as there is a workaround.

src/reportWebVitals.ts
Tulun: This comes default with CRA. We can remove it if need be. It helps with web reporting and such

tsconfig.json
Tulun: We can definitely dive into the documentation what each of these does -- this is just the default configuration from the template.

Gamaranto: The default looks good enough for now. We can change this later if we need more stuff or if we converge to some team wide ts rules.

package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "voter-dapp",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@emotion/core": "^11.0.0",
```
Gamaranto, line 6: I think `@emotion/core` is not necessary here as it should be included by `@emotion/styled` and `@emotion/react`

Gamaranto (approved):
Left some minor comments, but overall looks great. LGTM!

tailwind.config.js
```diff
@@ -0,0 +1,856 @@
+const colors = require('tailwindcss/colors')
```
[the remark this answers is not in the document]
Tulun: I don't mind either way. This was generated with: npx tailwindcss-cli@latest init --full.

We can pair it down to the empty version, or remove whatever.

daywiss (approved):
looks good

PR #4, opened 2021-03-30
refac(common): wrap onboard api into stateful class

daywiss:
experiment to see how we could break onboard out into its own class and have a common api for stateful classes in the form of:
```
  const Class = (initialState | config, depedencies:any, emitter: (state)=>void) => api:any

```
  which is easily consumed by a context or hook.


Tulun (commented):
Neat idea overall. I think this pattern would be potentially be good for libs, although it might not be suited for more complex global state interactions, where we might need something more robust.

PR #6, opened 2021-04-07
feat: Past requests renders with subgraph data

Tulun:
**Motivation**
Past Requests table renders out data from previous voting rounds so users can see previous votes, and if they voted correctly.

**Summary**
Using useVoteData hook which pulls data from subgraph. Data displays differently depending if they are logged in or not.

**Testing**

- [x]  New unit tests created
- [ ]  Untested


src/features/wallet/Wallet.tsx
```diff
@@ -0,0 +1,263 @@
...
+    font-weight: 600;
+    font-size: 1.5rem;
+  }
+  .sm-title {
+    color: #000;
+    margin: 0 0 12px;
+    font-weight: 400;
+    opacity: 0.5;
+  }
+  .value-tokens {
```
Gamaranto, line 155: same minor comments, classes can give specificity issues, as its easy to use the same class name in a parent and then the parent selector would have the same specificity (or more depending on the circumstances)

src/features/vote/hooks/useRewardsRetrievedEvents.ts
```diff
@@ -0,0 +1,31 @@
...
+    () => {
+      return queryRewardsRetrieved(contract, address).then((res) => {
+        if (res) {
+          return res;
+        } else {
+          return [];
+        }
+      });
+    },
+    { enabled: contract !== null ? true : false }
```
Gamaranto, line 23: I think you can get rid of the ternary here and just use `contract !=null`

Gamaranto (approved):
Some minor nits, but otherwise looks good!

src/web3/queryVotingContractEvents.ts
```diff
@@ -0,0 +1,337 @@
...
+      const datum = {} as PriceRound;
+      if (args) {
+        datum.roundId = args[0].toString();
+        datum.identifier = ethers.utils.toUtf8String(args[1]);
+        datum.time = args[2].toString();
+      }
+      return datum;
+    });
+  } catch (err) {
+    console.log("err", err);
```
daywiss, line 50: can we bubble these errors up rather than catching them? I think it would be more useful in the end to bubble all errors to caller as far as possible and handle errors there, wdyt?

src/web3/queryVotingContractEvents.ts
```diff
@@ -0,0 +1,337 @@
...
+}
+
+export const queryVotesCommitted = async (
+  contract: ethers.Contract | null,
+  address: string | null = null,
+  roundId: string | null = null,
+  identifier: string | null = null,
+  time: number | null = null
+) => {
+  if (!contract)
```
daywiss, line 77: would suggest using assert, its a bit cleaner syntax, though it can be unintuitive, its really `assert.ok`
```js
import assert from 'assert'
assert(contract,"User is not connected to provider and cannot query contract.")
```

src/features/vote/hooks/usePriceRoundEvents.ts
```diff
@@ -0,0 +1,34 @@
...
+const signer = provider.getSigner();
+const contract = createVotingContractInstance(signer);
+
+// This can be accessed without logging the user in.
+export default function usePriceRoundEvents() {
+  const { data, error, isFetching } = useQuery<PriceRound[]>(
+    "priceRoundEvents",
+    () => {
+      return queryPriceRoundEvents(contract).then((res) => {
+        if (res instanceof Error) return [];
```
daywiss, line 19: dont think res could be an error unless you are just returning an error rather than throwing, which would not be good.  I would suggest in these useQuery style calls, we just let the function throw errors and form a strategy to handle it holistically since useQuery provides an error state.  this would simplify your `then` handler, as you wont need to check for a null condition. we might have to come up with a convention for this, wdyt?

src/features/vote/hooks/usePriceRoundEvents.ts
```diff
@@ -0,0 +1,34 @@
...
+          return [];
+        }
+      });
+    }
+  );
+
+  if (data) {
+    return { priceRounds: data, error, isFetching };
+  } else {
+    return { priceRounds: [] as PriceRound[], error, isFetching };
```
daywiss, line 32: would be great if we didnt have to do this check before returning, and let the view check the the error /isFetching state to determine if it should use `priceRounds`.  The typescript would need to be changed to allow `undefined`. I'm also of the opinion that maybe we should just maintain the useQuery pattern and return any data as `data` and rename them in the view: `{data:priceRounds}`.   this would simplify a lot of these hooks:
```js

export default function usePriceRoundEvents() {
  return  useQuery<PriceRound[] | undefined>( "priceRoundEvents", () => {
      return queryPriceRoundEvents(contract)
    });
  );
}
```

 wdyt?

daywiss (approved):
looks great. really like how all the logic is broken out and react code ends up being really clean. 

i have some concerns around error handling, this could be improved. id much rather throw everything as far up the stack as possible, rather than catch it at the origin. i also think the hooks could be simplified this way by requiring views to handle data, loading and error states explicitly rather than hiding it in the hook. not a blocker in this PR but lets think about this some.

daywiss:
@Tulun check your DCO, you can sign off and force push the update, or sign off in github
image

src/web3/queryVotingContractMethods.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const queryRetrieveRewards = async (
+  contract: ethers.Contract,
+  address: string,
+  roundId: string,
+  identifier: string,
+  time: string
+) => {
+  try {
+    const reward: ethers.BigNumber = await contract.callStatic[
```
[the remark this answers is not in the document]
Tulun: I kept getting undefined until I added this name directly. I believe this is how the key is defined in the JSON.

src/features/vote/hooks/useRewardsRetrievedEvents.ts
Tulun: Good call.

src/features/wallet/Wallet.tsx
Tulun: Let me test this out.

Tulun: I think I might have a workaround that's not too intrusive.

It would mean you could still add stylings above, but could easily override them locally.

src/features/vote/hooks/usePriceRoundEvents.ts
Tulun: Totally agree. I forgot to change this one over.

src/web3/queryVotingContractEvents.ts
Tulun: Yeah that's a nice clean syntax.

PR #8, opened 2021-04-20
feat: retrieve all rewards works 

Tulun:
**Motivation**

Voter should be able to collect all rewards available to them after round ends.

**Summary**

User may click the "collect" button on the wallet section if rewards are available; this will trigger a request to their wallet.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/src/features/vote/ActiveRequestsForm.tsx
+++ b/src/features/vote/ActiveRequestsForm.tsx
@@ -229,7 +229,7 @@ const ActiveRequestsForm: FC<Props> = ({
         <thead>
           <tr>
             <th>Requested Vote</th>
-            <th>Proposal Detail</th>
+            {/* <th>Proposal Detail</th> */}
             <th>Description</th>
             <th>Commit Vote</th>
             <th>Your Vote</th>
@@ -243,7 +243,7 @@ const ActiveRequestsForm: FC<Props> = ({
                 <td>
                   <div className="identifier">{el.identifier}</div>
                 </td>
-                <td>{el.ancillaryData}</td>
+                {/* <td>{el.ancillaryData}</td> */}
                 <td>
                   <div className="description">
                     Lorem ipsum dolor sit amet, consectetur adipiscing elit.
--- a/src/features/vote/ActiveRequestsForm.tsx
+++ b/src/features/vote/ActiveRequestsForm.tsx
@@ -20,6 +20,7 @@ import { formatVoteDataToCommit } from "./helpers/formatVoteDataToCommit";
 import { EncryptedVote } from "web3/get/queryEncryptedVotesEvents";
 
 import { FormWrapper, ModalWrapper } from "./styled/ActiveRequestsForm.styled";
+import { VoteRevealed } from "web3/get/queryVotesRevealedEvents";
 
 export type FormData = {
   [key: string]: string;
@@ -37,12 +38,14 @@ interface Props {
   votePhase: string;
   encryptedVotes: EncryptedVote[];
   refetchEncryptedVotes: Function;
+  revealedVotes: VoteRevealed[];
 }
 
 interface TableValue {
   ancillaryData: string;
   identifier: string;
   vote: string;
+  revealed: boolean;
 }
 
 const ActiveRequestsForm: FC<Props> = ({
@@ -52,6 +55,7 @@ const ActiveRequestsForm: FC<Props> = ({
   votePhase,
   encryptedVotes,
   refetchEncryptedVotes,
+  revealedVotes,
 }) => {
   const [modalState, setModalState] = useState<
     "init" | "pending" | "success" | "error"
@@ -61,12 +65,28 @@ const ActiveRequestsForm: FC<Props> = ({
     state: { address, network, signer },
   } = useContext(OnboardContext);
   const { votingContract } = useVotingContract(signer, isConnected, network);
-  // const { data } = useVotesCommittedEvents(votingContract, address);
 
   const [tableValues, setTableValues] = useState<TableValue[]>([]);
 
   const { data: roundId } = useCurrentRoundId();
   const { isOpen, open, close, modalRef } = useModal();
+  const [canReveal, setCanReveal] = useState(false);
+
+  useEffect(() => {
+    if (encryptedVotes.length && votePhase === "Reveal") {
+      if (revealedVotes.length) {
+        revealedVotes.forEach((el) => {
+          const findRevealedVote = encryptedVotes.find(
+            (x) => x.identifier === el.identifier
+          );
+          // If there are no revealed votes and some encrypted votes, set can reveal to true.
+          if (!findRevealedVote) setCanReveal(true);
+        });
+      }
+    } else {
+      setCanReveal(false);
+    }
+  }, [encryptedVotes, votePhase, revealedVotes]);
 
   const generateDefaultValues = useCallback(() => {
     const dv = {} as FormData;
@@ -157,6 +177,7 @@ const ActiveRequestsForm: FC<Props> = ({
           ancillaryData: el.ancillaryData,
           vote: "-",
           identifier: el.identifier,
+          revealed: false,
         };
       });
 
@@ -183,11 +204,20 @@ const ActiveRequestsForm: FC<Props> = ({
         } else {
           datum.vote = vote;
         }
+        const findReveal = revealedVotes.find(
+          (x) => x.identifier === el.identifier
+        );
+        if (findReveal) {
+          datum.revealed = true;
+        } else {
+          datum.revealed = false;
+        }
+
         tv.push(datum);
       });
       setTableValues(tv);
     }
-  }, [activeRequests, encryptedVotes]);
+  }, [activeRequests, encryptedVotes, revealedVotes]);
 
   return (
     <FormWrapper
@@ -245,15 +275,16 @@ const ActiveRequestsForm: FC<Props> = ({
                 <td>
                   <div>
                     {/* <UnlockedIcon /> */}
-                    {votePhase === "Commit" && el.vote
+                    {votePhase === "Commit" && el.vote !== "-"
                       ? "Committed"
-                      : votePhase === "Commit" && !el.vote
+                      : votePhase === "Commit"
                       ? "Uncommitted"
                       : null}
-                    {votePhase === "Reveal" && el.vote
+                    {votePhase === "Reveal" && el.revealed
+                      ? "Revealed"
+                      : el.vote && !el.revealed
                       ? "Reveal"
-                      : /* this is incorrect -- just wanna hide this string for now. */
-                        null}
+                      : "Uncommitted"}
                   </div>
                 </td>
               </tr>
@@ -281,7 +312,7 @@ const ActiveRequestsForm: FC<Props> = ({
               Commit Votes
             </Button>
           ) : null}
-          {votePhase === "Reveal" ? (
+          {votePhase === "Reveal" && canReveal ? (
             <Button
               type="button"
               onClick={() => {
@@ -330,7 +361,11 @@ const ActiveRequestsForm: FC<Props> = ({
             >
               Reveal Votes
             </Button>
-          ) : null}
+          ) : (
+            <Button type="button" variant="disabled">
+              Reveal Votes
+            </Button>
+          )}
         </div>
       </div>
       <Modal isOpen={isOpen} onClose={close} ref={modalRef}>
--- a/src/features/vote/ActiveRequestsForm.tsx
+++ b/src/features/vote/ActiveRequestsForm.tsx
@@ -46,6 +46,7 @@ interface TableValue {
   identifier: string;
   vote: string;
   revealed: boolean;
+  ancHex: string;
 }
 
 const UNDEFINED_VOTE = "-";
@@ -185,6 +186,7 @@ const ActiveRequestsForm: FC<Props> = ({
           vote: "-",
           identifier: el.identifier,
           revealed: false,
+          ancHex: el.idenHex,
         };
       });
 
@@ -202,9 +204,10 @@ const ActiveRequestsForm: FC<Props> = ({
         const findVote = latestVotesFirst.find(
           (x) =>
             x.identifier === el.identifier &&
-            x.ancillaryData === el.ancillaryData &&
+            x.ancillaryData === el.ancHex &&
             x.time === el.time
         );
+
         if (findVote) {
           datum.vote = ethers.utils.formatEther(findVote.price);
           if (el.identifier.includes("Admin")) {
@@ -217,7 +220,7 @@ const ActiveRequestsForm: FC<Props> = ({
         const findReveal = revealedVotes.find(
           (x) =>
             x.identifier === el.identifier &&
-            x.ancillaryData === el.ancillaryData &&
+            x.ancillaryData === el.ancHex &&
             x.time === el.time
         );
         if (findReveal) {
--- a/src/features/vote/ActiveRequestsForm.tsx
+++ b/src/features/vote/ActiveRequestsForm.tsx
@@ -48,6 +48,8 @@ interface TableValue {
   revealed: boolean;
 }
 
+const UNDEFINED_VOTE = "-";
+
 const ActiveRequestsForm: FC<Props> = ({
   activeRequests,
   isConnected,
@@ -82,6 +84,8 @@ const ActiveRequestsForm: FC<Props> = ({
           // If there are no revealed votes and some encrypted votes, set can reveal to true.
           if (!findRevealedVote) setCanReveal(true);
         });
+      } else {
+        setCanReveal(true);
       }
     } else {
       setCanReveal(false);
@@ -275,16 +279,18 @@ const ActiveRequestsForm: FC<Props> = ({
                 <td>
                   <div>
                     {/* <UnlockedIcon /> */}
-                    {votePhase === "Commit" && el.vote !== "-"
+                    {votePhase === "Commit" && el.vote !== UNDEFINED_VOTE
                       ? "Committed"
                       : votePhase === "Commit"
                       ? "Uncommitted"
                       : null}
                     {votePhase === "Reveal" && el.revealed
                       ? "Revealed"
-                      : votePhase === "Reveal" && el.vote && !el.revealed
+                      : votePhase === "Reveal" &&
+                        el.vote !== UNDEFINED_VOTE &&
+                        !el.revealed
                       ? "Reveal"
-                      : votePhase === "Reveal" && !el.vote
+                      : votePhase === "Reveal" && el.vote === UNDEFINED_VOTE
                       ? "Uncommitted"
                       : null}
                   </div>
@@ -334,7 +340,7 @@ const ActiveRequestsForm: FC<Props> = ({
                       datum.ancillaryData = el.ancillaryData;
                       // anc data is set to - or N/A in UI if empty, convert back to 0x.
                       if (
-                        el.ancillaryData === "-" ||
+                        el.ancillaryData === UNDEFINED_VOTE ||
                         el.ancillaryData === "N/A"
                       ) {
                         datum.ancillaryData = "0x";
--- a/src/features/vote/ActiveRequestsForm.tsx
+++ b/src/features/vote/ActiveRequestsForm.tsx
@@ -282,9 +282,11 @@ const ActiveRequestsForm: FC<Props> = ({
                       : null}
                     {votePhase === "Reveal" && el.revealed
                       ? "Revealed"
-                      : el.vote && !el.revealed
+                      : votePhase === "Reveal" && el.vote && !el.revealed
                       ? "Reveal"
-                      : "Uncommitted"}
+                      : votePhase === "Reveal" && !el.vote
+                      ? "Uncommitted"
+                      : null}
                   </div>
                 </td>
               </tr>
@@ -361,11 +363,11 @@ const ActiveRequestsForm: FC<Props> = ({
             >
               Reveal Votes
             </Button>
-          ) : (
+          ) : votePhase === "Reveal" && !canReveal ? (
             <Button type="button" variant="disabled">
               Reveal Votes
             </Button>
-          )}
+          ) : null}
         </div>
       </div>
       <Modal isOpen={isOpen} onClose={close} ref={modalRef}>
--- a/src/features/vote/ActiveRequestsForm.tsx
+++ b/src/features/vote/ActiveRequestsForm.tsx
@@ -233,6 +233,7 @@ const ActiveRequestsForm: FC<Props> = ({
         <thead>
           <tr>
             <th>Requested Vote</th>
+            {/* Commented out for now -- might move the anc data elsewhere */}
             {/* <th>Proposal Detail</th> */}
             <th>Description</th>
             <th>Commit Vote</th>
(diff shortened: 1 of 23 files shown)


src/web3/post/retrieveRewards.ts
```diff
@@ -0,0 +1,42 @@
...
+) => {
+  try {
+    const tx = await contract.functions[
+      "retrieveRewards(address,uint256,(bytes32,uint256,bytes)[])"
+    ](data.voterAddress, data.roundId, data.pendingRequests);
+
+    console.log("retrieve rewards TX?", tx);
+    return tx;
+  } catch (err) {
+    console.log("Err in attempted retrieveRewards transaction", err);
```
daywiss, line 40: some logs left in here. might want to just throw error

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -0,0 +1,37 @@
...
+    pendingRequestData.push(pendingRequest);
+  });
+
+  postData.pendingRequests = pendingRequestData;
+  retrieveRewards(contract, postData).then((tx) => {
+    tx.wait(1).then((conf: any) => {
+      // This function should collect all available rewards. Set balance to 0.
+      // Note: This still needs to be tested for N-1, N-2, ... rounds.
+      setAvailableRewards(DEFAULT_BALANCE);
+    });
```
daywiss, line 35: just a nit here, but you can return the promise and unwind the nesting. also the call should be returned or caught if not returned
```js
  // this should be returned so we can catch error upstream
  return retrieveRewards(contract, postData)
     .then((tx) => tx.wait(1))
     .then((conf: any) => {
        // This function should collect all available rewards. Set balance to 0.
        // Note: This still needs to be tested for N-1, N-2, ... rounds.
        setAvailableRewards(DEFAULT_BALANCE);
     })
```

src/features/wallet/helpers/checkAvailableRewards.ts
```diff
@@ -0,0 +1,30 @@
...
+) {
+  const promises = data.map(async (vote) => {
+    const rewardAvailable = await queryRetrieveRewards(
+      contract,
+      address,
+      vote.roundId,
+      vote.identifier,
+      vote.time
+    );
+    if (rewardAvailable) return parseFloat(rewardAvailable);
```
daywiss, line 18: hopefully this doesnt affect reward getting, and only display, but this can cause some information loss / dust if value is used for claiming (which it doesnt look like it does).

src/features/wallet/helpers/checkAvailableRewards.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  return Promise.all(promises).then((values) => {
+    let balance = 0;
+    values.map((val) => {
+      if (val && val) return (balance += val);
```
daywiss, line 24: is `val && val` a typo?

src/features/wallet/helpers/checkAvailableRewards.ts
```diff
@@ -0,0 +1,30 @@
...
+      return false;
+    });
```
daywiss, line 26: im surprised typescript is ok with this, returning bool or number.  just a nit but id probably recommend doing this as a reduce operation
```js
const values = await Promise.all(promises)
return values.reduce((balance,val)=>{
  if (val && val) return balance + val;
  return balance
},0)

```

src/features/wallet/helpers/checkAvailableRewards.ts
```diff
@@ -0,0 +1,30 @@
...
+      contract,
+      address,
+      vote.roundId,
+      vote.identifier,
+      vote.time
+    );
+    if (rewardAvailable) return parseFloat(rewardAvailable);
+  });
+
+  return Promise.all(promises).then((values) => {
```
daywiss, line 21: why not use await

src/features/wallet/helpers/calculateUMATotalValue.ts
```diff
@@ -0,0 +1,9 @@
+export default function calculateUMATotalValue(price: number, balance: string) {
```
daywiss, line 1: this is cool, but just fyi ethers does have this utility [commify](https://docs.ethers.io/v5/api/utils/display-logic/#utils-commify), although your implementation is nice, so this is just an fyi

daywiss (approved):
looks good to me, thanks for keeping this pr small. did not really get to test the reveal stuff, but the wallet stats looked good.

src/features/wallet/helpers/collectRewards.ts
Tulun: Good call, that's cleaner. I think we may want to do some pending state stuff, but that's not currently happening.

src/features/wallet/helpers/checkAvailableRewards.ts
Tulun: Yeah this is just for the balance in the 3rd column. It should line up for actual values available, but it could be mismatched. But it won't matter -- the actual blockchain call will determine that.

src/features/wallet/helpers/checkAvailableRewards.ts
Tulun: Yeah I was trying to do promise.allSettled and forgot to change out.

src/features/wallet/helpers/calculateUMATotalValue.ts
Tulun: Good to know, figured ethers didn't have something like that.

src/features/wallet/helpers/checkAvailableRewards.ts
Tulun: I refactored it to this, a balance can be undefined (promise):

  return values.reduce((balance, val) => {
    if (val) return balance ? balance : 0 + val;
    return balance;
  }, 0);

src/features/vote/ActiveRequestsForm.tsx
```diff
@@ -199,7 +233,7 @@ const ActiveRequestsForm: FC<Props> = ({
         <thead>
           <tr>
             <th>Requested Vote</th>
-            <th>Proposal Detail</th>
+            {/* <th>Proposal Detail</th> */}
```
mrice32, line 236: Why is this commented out?

If it's meant to stay commented, might be useful to add a note explaining why.

src/features/vote/ActiveRequestsForm.tsx
```diff
@@ -213,7 +247,7 @@ const ActiveRequestsForm: FC<Props> = ({
                 <td>
                   <div className="identifier">{el.identifier}</div>
                 </td>
-                <td>{el.ancillaryData}</td>
+                {/* <td>{el.ancillaryData}</td> */}
```
mrice32, line 250: Same as above.

src/features/vote/ActiveRequestsForm.tsx
```diff
@@ -61,12 +67,30 @@ const ActiveRequestsForm: FC<Props> = ({
...
   const { data: roundId } = useCurrentRoundId();
   const { isOpen, open, close, modalRef } = useModal();
+  const [canReveal, setCanReveal] = useState(false);
+
+  useEffect(() => {
+    if (encryptedVotes.length && votePhase === "Reveal") {
+      if (revealedVotes.length) {
+        revealedVotes.forEach((el) => {
+          const findRevealedVote = encryptedVotes.find(
+            (x) => x.identifier === el.identifier
```
mrice32, line 82: Will this cause false-positives (and thus cause set reveal to be false when it should be true)? What happens if there are multiple votes in a round with the same identifier, but different ancillary data and/or timestamp?

Don't we need to compare ancillary data and timestamp too?

src/features/vote/ActiveRequestsForm.tsx
```diff
@@ -183,11 +208,20 @@ const ActiveRequestsForm: FC<Props> = ({
         } else {
           datum.vote = vote;
         }
+        const findReveal = revealedVotes.find(
+          (x) => x.identifier === el.identifier
+        );
```
mrice32, line 213: Same question as above.

src/features/vote/ActiveRequestsForm.tsx
```diff
@@ -20,6 +20,7 @@ import { formatVoteDataToCommit } from "./helpers/formatVoteDataToCommit";
 import { EncryptedVote } from "web3/get/queryEncryptedVotesEvents";
 
 import { FormWrapper, ModalWrapper } from "./styled/ActiveRequestsForm.styled";
+import { VoteRevealed } from "web3/get/queryVotesRevealedEvents";
```
mrice32, line 23: nit: are any of the changes in this file related to the retrieve rewards feature? This isn't a show-stopper on this PR, but in the interest of reducing PR sizes, maybe changes like this should be put into a separate PR in the future?

Thoughts? Or am I misunderstanding the purpose of the changes in this file?

src/features/vote/ActiveRequestsForm.tsx
```diff
@@ -245,15 +279,20 @@ const ActiveRequestsForm: FC<Props> = ({
...
+                      ? "Revealed"
+                      : votePhase === "Reveal" &&
+                        el.vote !== UNDEFINED_VOTE &&
+                        !el.revealed
                       ? "Reveal"
-                      : /* this is incorrect -- just wanna hide this string for now. */
-                        null}
+                      : votePhase === "Reveal" && el.vote === UNDEFINED_VOTE
+                      ? "Uncommitted"
+                      : null}
```
mrice32, line 295: Thoughts on an alternative approach where you do the following above the JSX:
```js
tableValues.map((el, index) => {
  let stateEntry;
  let voteEntry;
  switch(votePhase) {
    case "Commit":
      // Logic
      break;
    case "Reaveal":
      // Logic
      break;
  }
}
// .. return JSX down here
```

There's certainly a tradeoff here -- the logic is placed away from where it's displayed. I only raise the point because nested ternaries are pretty hard to follow, but there doesn't seem to be a good alternative when the logic is inside the JSX.

mrice32 (commented):
Looks awesome! Some initial comments. Will continue the review shortly!

src/features/vote/ActiveRequestsForm.tsx
Tulun: Oh yes, sorry, this is because we are hiding the anc data column for now. I can add a comment np, or just delete it. It's a trivial thing to add back in.

src/features/vote/ActiveRequestsForm.tsx
Tulun: Same as above ;)

src/features/vote/ActiveRequestsForm.tsx
Tulun: I think the data filters based on the current round, but I could verify that with a bigger data set. If so, you would just need to add those comparisons like you said.

src/features/vote/ActiveRequestsForm.tsx
Tulun: Oh this was largely finishing off some work that was tagged onto my larger previous PR. as it was a rather small addition I didn't feel the need to do a PR separate

Highly supportive of smaller PRs, though, we just need a fast turn around.

src/features/vote/ActiveRequestsForm.tsx
Tulun: Yeah it's more a stylistic choice.

To be honest, because of how complex this ActiveRequest form view *is* in general, it is tempting to split off the commit and reveal states into their own views in general. There are potentially enough differences between the views on my chatting with Tom that this is probably okay. To be honest, the "reveal" stage, this is actually no longer a form, not in the sense we'd think about in an HTML sense.

src/features/vote/ActiveRequestsForm.tsx
mrice32: Up to you -- I think either an explanatory note or deleting SGTM!

src/features/vote/ActiveRequestsForm.tsx
mrice32: Yeah, I would argue that this is the exact sort of change that warrants its own PR. I think smaller PRs often _help_ with fast turnaround bc reviewers tend to procrastinate on PRs that are 100+ lines in my experience.

src/features/vote/ActiveRequestsForm.tsx
mrice32: Yep, I'm saying I think it needs to compare ancillaryData and timestamp in addition to the identifier. And yeah, agreed that roundId comparison shouldn't be necessary.

src/features/vote/ActiveRequestsForm.tsx
mrice32: Agreed on all of the above. I'm not super strongly opinionated. My gut response to seeing nested ternaries is that it's hard to read. But I totally get the constraints of JSX. Just wanted to mention it, but seems like you've already given it a decent amount of thought.

src/features/vote/ActiveRequestsForm.tsx
Tulun: Sounds good! I'll add it in.

src/features/wallet/helpers/calculateUMATotalValue.ts
mrice32: Any reason not to use that instead of the regex as a matter of simplicity?

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -0,0 +1,42 @@
...
+  PostRetrieveReward,
+  PendingRequestRetrieveReward,
+} from "web3/post/retrieveRewards";
+
+const DEFAULT_BALANCE = "0";
+
+export default function collectRewards(
+  contract: ethers.Contract,
+  data: RewardsRetrieved[],
+  setAvailableRewards: Function
```
mrice32, line 14: Can this type be more specific?

src/features/wallet/helpers/checkAvailableRewards.ts
```diff
@@ -0,0 +1,27 @@
...
+    );
+    if (rewardAvailable) return parseFloat(rewardAvailable);
+  });
+
+  const values = await Promise.all(promises);
+
+  return values.reduce((balance, val) => {
+    if (val) return (balance ? balance : 0) + val;
+    return balance;
+  }, 0);
```
mrice32, line 26: Can this be simplified or am I missing something? I assume you should be able to tell typescript that balance _must_ be defined, right?
```suggestion
  return values.reduce((balance, val) => balance + val ? val : 0, 0);
```

src/features/wallet/helpers/calculateUMATotalValue.ts
Tulun: Nope! That works just fine. 

Fixed, with an improvement for the 0 balance case as well.

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -0,0 +1,42 @@
...
+  postData.pendingRequests = pendingRequestData;
+  // this should be returned so we can catch error upstream
+  return (
+    retrieveRewards(contract, postData)
+      // wait for at least 1 block conf.
+      .then((tx) => tx.wait(1))
+      .then((conf: any) => {
+        // This function should collect all available rewards. Set balance to 0.
+        // Note: This still needs to be tested for N-1, N-2, ... rounds.
+        setAvailableRewards(DEFAULT_BALANCE);
```
mrice32, line 38: This shouldn't necessarily collect for _all_ rounds, right? In fact, if you haven't claimed for multiple rounds, it _can't_ collect all rewards, right?

src/features/wallet/helpers/formatWalletBalance.ts
```diff
@@ -0,0 +1,28 @@
+// There are two different colours for the first 4 digits and last 4 digits of the number.
+// Hence we need to return an array of strings.
+// We want to limit the overall string length to ~8 digits.
+export default function formatWalletBalance(balance: string): string[] {
```
mrice32, line 4: Similar to my other comment above, I think this might be better suited for a different PR unless this deals with rewards (although, now that it's here, I think it's fine to keep in).

src/features/wallet/helpers/checkAvailableRewards.ts
```diff
@@ -0,0 +1,27 @@
...
+import { VoteRevealed } from "web3/get/queryVotesRevealedEvents";
+import { queryRetrieveRewards } from "web3/get/queryRetrieveRewards";
+
+export default async function checkAvailableRewards(
+  data: VoteRevealed[],
+  address: string,
+  contract: ethers.Contract
+) {
+  const promises = data.map(async (vote) => {
+    const rewardAvailable = await queryRetrieveRewards(
```
mrice32, line 11: Minor note (not sure if there's anything to change): it looks like we're casting this to a float then using it to determine _if_ there are any rewards to claim. If we have rewards that are so small that they are below the precision limit of JS floats, we may accidentally truncate nonzero rewards to 0 and think that the user has _no_ rewards to claim when they really have just a very small number. I'm not sure how small their initial balance would need to be for this to happen, but just wanted to mention it since this float seems to be used for more than just display.

mrice32 (commented):
A few more comments, but claim logic generally looks quite clean!

src/features/wallet/helpers/collectRewards.ts
Tulun: setAvailableRewards you mean?

src/features/wallet/helpers/checkAvailableRewards.ts
daywiss: i noticed this as well, and you bring up a really good edge case. I think in general we should default to handling these numbers as wei strings/bn's until its time to display / math them. and maybe define some guidelines around how to deal with BNs across all our apps since its tricky.

src/web3/get/queryRetrieveRewards.ts
```diff
@@ -1,22 +1,40 @@
...
+      {
+        identifier: stringToBytes32(identifier),
+        time: Number(time),
+        ancillaryData,
+      },
     ]);
 
     return ethers.utils.formatEther(reward.toString());
   } catch (err) {
+    console.log("identifier", identifier);
```
mrice32, line 36: Is this console log meant to be left in here?

src/web3/createMulticallContractInstance.ts
```diff
@@ -0,0 +1,35 @@
+import MulticallAtrifact from "@uma/core/build/contracts/MulticallMock.json";
```
mrice32, line 1: nit: probably slightly preferable to import the interface `Multicall` rather than `MulticallMock` (the implementation) since you shouldn't need the full implementation to know how to interact with it.

src/web3/createMulticallContractInstance.ts
```diff
@@ -0,0 +1,35 @@
...
+  // const artifact: Network = MulticallAtrifact.networks;
+  // const network = artifact[networkId];
+  // console.log("network", network, "artifact", artifact, MulticallAtrifact);
+
+  const contract = new ethers.Contract(
+    // network.address,
+    // This is stubbed in for now. We need to develop this formally for mainnet / kovan.
+    process.env.REACT_APP_CURRENT_ENV === "test"
+      ? "0xD6b61cAd80DBe2Fe26Fa672b8Ec2FdAf002CeBbC"
+      : "",
```
mrice32, line 29: nit: you can pull in pre-deployed addresses by network id by either importing the makerdao file that lists them via their package:
```
import multicallAddresses from "@makerdao/multicall/src/addresses";

const networkName = "kovan";
const multicallAddress = multicallAddresses[networkName];
```

OR what might be easier is just to copy the [mapping](https://github.com/makerdao/multicall.js/blob/master/src/addresses.json) into this file:
```
{
  "mainnet": {
    "multicall": "0xeefba1e63905ef1d7acba5a8513c70307c1ce441",
    "rpcUrl": "https://mainnet.infura.io"
  },
  "kovan": {
    "multicall": "0x2cc8688c5f75e365aaeeb4ea8d6a480405a48d2a",
    "rpcUrl": "https://kovan.infura.io"
  },
  "rinkeby": {
    "multicall": "0x42ad527de7d4e9d9d011ac45b31d8551f8fe9821",
    "rpcUrl": "https://rinkeby.infura.io"
  },
  "goerli": {
    "multicall": "0x77dca2c955b15e9de4dbbcf1246b4b85b651e50e",
    "rpcUrl": "https://rpc.slock.it/goerli"
  },
  "xdai": {
    "multicall": "0xb5b692a88bdfc81ca69dcb1d924f59f0413a602a",
    "rpcUrl": "https://dai.poa.network"
  }
}
```

You could remove the RPC urls and change the names to network IDs to make it easier to look up at runtime.

src/hooks/useMulticall.ts
```diff
@@ -0,0 +1,30 @@
...
+  ] = useState<ethers.Contract | null>(null);
+
+  useEffect(() => {
+    // If connected, try to create contract with assigned signer.
+    if (isConnected) {
+      // Signer can be null check for null and if we've already defined a contract.
+      if (signer && !multicallContract && network) {
+        const contract = createMulticallContractInstance(
+          signer
+          // network.chainId.toString()
```
mrice32, line 22: Commented out?

src/web3/createMulticallContractInstance.ts
```diff
@@ -0,0 +1,35 @@
...
+import { ethers } from "ethers";
+
+// interface Network {
+//   [key: string]: {
+//     address: string;
+//     events: object;
+//     links: object;
+//     transactionHash: string;
+//   };
+// }
```
mrice32, line 11: Commented out?

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -0,0 +1,76 @@
...
+  setAvailableRewards: Dispatch<SetStateAction<string>>,
+  multicallContract: ethers.Contract
+) {
+  const multicallCollectRewards = [] as MulticallCollectRewards[];
+  const votingInterface = new ethers.utils.Interface(VotingArtifact.abi);
+
+  // find the unique roundIds
+  const uniqueRoundIds = data
+    .map((item) => item.roundId)
+    .filter((value, index, self) => self.indexOf(value) === index);
```
mrice32, line 32: I really wish that the array had a native `.unique()` method rather than having to implement it yourself or use lodash, so this wouldn't be required to do something so simple.

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const DEFAULT_BALANCE = "0";
+
+interface MulticallCollectRewards {
+  // contract address
+  target: string;
+  callData: string;
+}
+
+export default function collectRewards(
```
mrice32, line 19: +1 this function looks great!

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -0,0 +1,76 @@
...
+  PostRetrieveReward,
+  PendingRequestRetrieveReward,
+} from "web3/post/retrieveRewards";
+
+import VotingArtifact from "@uma/core/build/contracts/Voting.json";
+
+const DEFAULT_BALANCE = "0";
+
+interface MulticallCollectRewards {
+  // contract address
```
mrice32, line 14: comment?

mrice32 (approved):
LGTM! Only a few minor comments around how we're pulling in multicall.

src/web3/createMulticallContractInstance.ts
Tulun: Good to know. If there are already kovan / mainnet deployments out there, we just need to worry about the locally deployed ones.

src/web3/createMulticallContractInstance.ts
Tulun: Only used if the networks object is defined. I wasn't sure if this would be the case, but it's looking like we aren't going to be using artifacts anyway.

PR #9, opened 2021-04-27
Feat: Past requests only uses thegraph; Summary details works.

Tulun:
**Motivation**
Past requests was refactored to only use thegraph; Summary details built.

**Summary**

Past Requests now only had READ data and does 0 contract interaction -- the wallet individual claims will be moved into the wallet section of the app.

Summary stats for each price request have been fleshed out into a modal.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


src/features/vote/PastRequests.tsx
```diff
@@ -60,15 +85,33 @@ const PastRequests: FC<Props> = ({ priceRounds, address, contract }) => {
...
-                    <div className="identifier">{el.proposal}</div>
+                    <div className="identifier">
+                      <p>{el.proposal}</p>
+                      <p
+                        onClick={() => {
+                          open();
+                          setModalState({
+                            proposal: el.proposal,
+                            correct: el.correct,
+                            totalSupply: el.totalSupply,
```
Gamaranto, line 103: this line crashes for me when clicking on "View Details"

Gamaranto (approved):
Get a crash when clicking on "view details" , but I suspect it might just be an env variables misconfiguration, otherwise LGTM

src/features/vote/PastRequests.tsx
daywiss: this fails on the most recent past requests, admin 86 and admin 87. probably need to set some kind of default values.

daywiss (changes requested):
everything looks good to me code wise. testing the app there is a crash which francesco mentions. also i think maybe we should clean up these decimals somehow, its just excessive.

image

Tulun:
> everything looks good to me code wise. testing the app there is a crash which francesco mentions. also i think maybe we should clean up these decimals somehow, its just excessive.
> 
> image

lol np. I know you and Matt were giving me flak for not displaying the values, but yeah, let's truncate it down to 6 or something?

src/features/vote/PastRequests.tsx
Tulun: These shouldn't even be displayed. Lemme filter them out -- they aren't technically Past Requests, and so don't have certain pieces of data.

I should probably do some kind of catch though, in case the data hasn't been properly indexed yet.

daywiss (approved):
looks great

PR #14, opened 2021-05-07
fix: clear keys on account switch, decode only newest commits

daywiss:
This has 3 fixes:
1. clears private/public keys when new accounts are detected
2. ensures react query does not cache encrypted data between accounts
3. de-duplicates commit events to ensure only the latest events get decrypted

### DIFF
--- a/src/web3/get/queryEncryptedVotesEvents.ts
+++ b/src/web3/get/queryEncryptedVotesEvents.ts
@@ -49,20 +49,28 @@ export const queryEncryptedVotes = async (
 
   try {
     const events = await contract.queryFilter(filter, VOTER_CONTRACT_BLOCK);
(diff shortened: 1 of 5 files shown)


src/web3/get/queryEncryptedVotesEvents.ts
```diff
@@ -49,20 +49,28 @@ export const queryEncryptedVotes = async (
 
   try {
     const events = await contract.queryFilter(filter, VOTER_CONTRACT_BLOCK);
+    // there may be multiple commit events, if there are collisions, then take newest
+    const eventTable = events.reduce((result:{[key:string]:any},event:any)=>{
```
daywiss, line 53: i dont know how to get the right types here

Tulun (approved):

PR #16, opened 2021-05-10
feat(voter-dapp): add CMS integration

Gamaranto:
**Motivation**

This PR integrates the Voter Dapp with Contentful and makes the View Past Requests Modal render the actual description of the UMIP as set in Contentful. I've added 2 UMIPs to Contentful (88, 89), so its expected that the others past requests don't render anything in the description field.

**Summary**

Briefly summarize what changes were made to accomplish the motivation above.

**Details**

This may be unnecessary for some PRs. Catch-all for detailed explanations about the implementation decisions and implications of the change.

**Testing**

- [ ]  New unit tests created
- [x]  Existing tests adequate, no new tests required
- [ ]  All existing tests pass
- [ ]  Untested


Tulun (approved):
LGTM.

I normally like breaking up the stuff in useUMIP, but its so self contained anyway I think this is fine.

src/features/vote/PastViewDetailsModal.tsx
```diff
@@ -87,21 +93,16 @@ const _PastViewDetailsModal: ForwardRefRenderFunction<
...
-            vehicula molestie sapien, eu dapibus metus auctor in. Nullam viverra
-            urna odio, sit amet lobortis metus interdum id. Nulla enim justo,
-            eleifend in metus in, eleifend dapibus ante. Donec nec egestas
-            lacus.
+            <ReactMarkdown
+              components={{
+                h1: MiniHeader,
+              }}
+            >
+              {umip?.description}
```
daywiss, line 104: Should this have some kind of place holder, like "No description" or will it look fine being empty

src/hooks/useUMIP.ts
```diff
@@ -0,0 +1,39 @@
...
+      enabled: number != null,
+    }
+  );
+
+  return { umip: data, ...others };
+}
+
+const contenfulSpaceId = process.env.REACT_APP_PUBLIC_CONTENTFUL_SPACE_ID;
+const contentfulAccessToken =
+  process.env.REACT_APP_PUBLIC_CONTENTFUL_ACCESS_TOKEN;
```
daywiss, line 18: can we update readme with this info?

src/hooks/useUMIP.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const contentfulClient = contentful.createClient({
+  space: contenfulSpaceId || "",
+  accessToken: contentfulAccessToken || "",
```
daywiss, line 22: i think contentful should be optional, especially during dev/staging.  this is technically optional, but things might break having empty strings here.  i have a suggestion for this let me know what u think

daywiss (commented):
this currently breaks the preview in vercel, is there a way to keep it working? might be missing some envs.  I have a suggestion im comments that could make this optional by requiring an env
`REACT_APP_PUBLIC_CONTENTFUL_ENABLED` to make this like a feature flag

src/features/vote/PastViewDetailsModal.tsx
Gamaranto: Not use, but someone from the content team will

src/hooks/useUMIP.ts
Gamaranto: Done!

src/hooks/useUMIP.ts
Gamaranto: Could you elaborate on why you think it should be optional? I think we would want to develop in an env as close as possible to the production one, to catch things as early as possible.
My only worry is if something breaks on the contenful integration the we would find out in prod if we don't use in development

src/features/vote/PastViewDetailsModal.tsx
Gamaranto: Also, I've added a fallback in case there is no description found.

src/hooks/useUMIP.ts
daywiss: well for this particular app, i dont see how we can do this since we wont always be testing on mainnet. how would contentful work on a testnet or kovan or something like that?

Gamaranto: Oh I see what you mean now, Contentful would still work, but the ids in the CMS would not match the actual requests done in the `dev` environment. For example if you added a proposal `Admin 100` on Kovan and the CMS had data for UMIP 100 then it would show the UMIP 100 description in the `dev` environment, conversely, if the CMS had no data for UMIP 100 then it would just show the fallback description

daywiss (approved):
looks good thanks

PR #18, opened 2021-05-12
Wallet two key select fixes

Tulun:
**Motivation**

Add in cold wallet address input.
**Summary**

Add in the cold wallet address input. 3rd wallet is required for two key contract for withdrawal. Previously it was the msg.sender, which although works, is not the intention of the two key contract.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/src/features/wallet/TwoKeyContractModal.tsx
+++ b/src/features/wallet/TwoKeyContractModal.tsx
@@ -41,12 +41,13 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
   const [showForm, setShowForm] = useState(false);
   const [value, setValue] = useState("");
   const [error, setError] = useState("");
-
+  const [success, setSuccess] = useState(false);
   const closeForm = useCallback(() => {
     if (showForm) {
       setShowForm(false);
       setValue("");
     } else {
+      setSuccess(false);
       setShowForm(true);
     }
   }, [showForm]);
@@ -59,14 +60,21 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
       return createDesignatedVotingContract(value, signer, network).then(
         (res) => {
           console.log("Success dvc?", res);
-          setValue("");
-          setError("");
+          closeForm();
+          setSuccess(true);
         }
       );
     }
-  }, [value, network, signer]);
+  }, [value, network, signer, closeForm]);
   return (
-    <Modal isOpen={isOpen} onClose={close} ref={externalRef}>
+    <Modal
+      isOpen={isOpen}
+      onClose={() => {
+        close();
+        setSuccess(false);
+      }}
+      ref={externalRef}
+    >
       <ModalWrapper>
         <h3 className="header">Two Key Voting</h3>
         {!isConnected ? (
@@ -103,6 +111,7 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
             </div>
           </>
         )}
+        {success ? <div>Successfully added two-key contract.</div> : null}
         {showForm ? (
           <FormWrapper>
             <StyledInput>
--- a/src/features/wallet/TwoKeyContractModal.tsx
+++ b/src/features/wallet/TwoKeyContractModal.tsx
@@ -104,7 +104,7 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
                     className="open-form"
                     tw="flex-grow text-right"
                   >
-                    Add Cold Wallet Address
+                    {showForm ? "Remove" : "Add Cold Wallet Address"}
                   </div>
                 </>
               ) : (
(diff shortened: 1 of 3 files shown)


src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -24,8 +38,43 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
...
+  }, [showForm]);
+
+  const submitForm = useCallback(() => {
+    if (value.substr(0, 2).toLowerCase() !== "0x")
+      return setError("Address must start with 0x");
+    if (value.length !== 42) return setError("Address must be 42 characters");
+    if (value && network && signer) {
+      return createDesignatedVotingContract(value, signer, network).then(
+        (res) => {
+          console.log("Success dvc?", res);
```
daywiss, line 62: prob dont need this

src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -48,20 +97,12 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
...
-                      if (votingAddress && network && signer) {
-                        createDesignatedVotingContract(
-                          votingAddress,
-                          signer,
-                          network
-                        ).then((res) => {
-                          console.log("Success dvc?", res);
-                        });
-                      }
+                      closeForm();
```
daywiss, line 100: small nit but why not just pass this function directly into the handler `onClick={closeForm}`

src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -24,8 +38,43 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
...
+  const [value, setValue] = useState("");
+  const [error, setError] = useState("");
+  const [success, setSuccess] = useState(false);
+  const closeForm = useCallback(() => {
+    if (showForm) {
+      setShowForm(false);
+      setValue("");
+    } else {
+      setSuccess(false);
+      setShowForm(true);
```
daywiss, line 51: is `closeForm` meant to toggle the form on and off?

src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -24,8 +38,43 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
...
+    }
+  }, [showForm]);
+
+  const submitForm = useCallback(() => {
+    if (value.substr(0, 2).toLowerCase() !== "0x")
+      return setError("Address must start with 0x");
+    if (value.length !== 42) return setError("Address must be 42 characters");
+    if (value && network && signer) {
+      return createDesignatedVotingContract(value, signer, network).then(
```
daywiss, line 60: would highly recommend catching an error and doing something with it, even if its just to log

src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -70,6 +111,35 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
             </div>
           </>
         )}
+        {success ? <div>Successfully added two-key contract.</div> : null}
+        {showForm ? (
```
daywiss, line 115: small nit, but if you want to show an element if true, and nothing if false i typically do this because its slightly shorter:
` success && <div>Successfully added two-key contract.</div>`

src/features/wallet/TwoKeyContractModal.tsx
Tulun: For which onClick handler?

src/features/wallet/TwoKeyContractModal.tsx
Tulun: Yeah if you look at the design, you open the form through what  looks like a anchor tag

https://old-star.surge.sh/#

src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -24,8 +38,43 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
...
+    } else {
+      setSuccess(false);
+      setShowForm(true);
+    }
+  }, [showForm]);
+
+  const submitForm = useCallback(() => {
+    if (value.substr(0, 2).toLowerCase() !== "0x")
+      return setError("Address must start with 0x");
+    if (value.length !== 42) return setError("Address must be 42 characters");
```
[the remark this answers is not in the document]
Tulun: Yeah that's nice, although we are being pickier about whether or not they have the prefix currently. Do you think we should add the 0x if the user omits it to the value? Or does the contract accept a 40 character address and just prepends the 0x anyway?

src/features/wallet/TwoKeyContractModal.tsx
Tulun: Good call.

src/features/wallet/TwoKeyContractModal.tsx
Tulun: Yup, no issue doing it this way.

src/features/wallet/TwoKeyContractModal.tsx
Tulun: Ok!

daywiss (approved):

PR #19, opened 2021-05-13
Feat: Select UI component

Tulun:
**Motivation**
Create dropdown view using Downshift.

**Summary**

Create a customized select view using downshift js. Created a non-react-hook-form and react-hook-form compatible component.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


src/common/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+            <FontAwesomeIcon icon={faAngleDown} />
+          </DownArrow>
+        )}
+      </DropdownHeader>
+      <DropdownList {...getMenuProps()} isOpen={isOpen}>
+        {isOpen &&
+          items.map((item, index) => (
+            <DropdownListItem
+              isHighlighted={highlightedIndex === index}
+              key={`${item.value}${index}`}
```
daywiss, line 54: im not sure if this is a good key since an item could change position in list, and then be considered a different item. is that ok?  theres also an edge case that this concatenation can lead to an ambiguous key.  wdyt?

src/common/components/dropdown/RHFDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+interface OptionType {
+  value: string;
+  label: string;
+}
+
+interface Props {
+  name: string;
+  items: OptionType[];
+  control: Control;
```
daywiss, line 23: just curious, this `control` prop isnt used, what is it meant for?

src/features/wallet/TwoKeyContractModal.tsx
```diff
@@ -88,11 +89,25 @@ const _TwoKeyContractModal: ForwardRefRenderFunction<
...
             <p tw="opacity-50 mb-4 text-center">
-              {!hotAddress
-                ? `You are not currently using a two key voting system. To deploy one,
-          provide your cold key address. Click here to learn more about the two
-          key voting system.`
-                : `Your hot wallet address is: ${hotAddress} and your cold wallet address is ${votingAddress}`}
+              {!hotAddress ? (
+                <div>
+                  You are not currently using a two key voting system. To deploy
+                  one, provide your cold key address. Click{" "}
```
daywiss, line 95: we should change the copy here and make it clear that this is optional @Tommy1231232

daywiss (commented):
i really like the composition of components here.  hopefully we can get some re-use out of this

src/features/wallet/TwoKeyContractModal.tsx
Tommy1231232: @daywiss I don't completely follow. What do you mean change the copy?

src/common/components/dropdown/RHFDropdown.tsx
Tulun: I'm not 100% sure, but react-hook-form throws a type error if it's not defined in the props.

src/common/components/dropdown/Dropdown.tsx
Tulun: We could do item.value alone? As the value of a dropdown should be unique.

daywiss: i think so

src/features/wallet/TwoKeyContractModal.tsx
daywiss: @Tommy1231232 by "copy" i just mean the text that is shown to the user. I think there was an issue with the old app that it wasnt clear this was optional and people waste gas making a 2 key contract. this doesnt give the user any indication that this is optional, and they probably dont want to do it unless they were told to

Tommy1231232: Gotcha, good idea I'll make a separate CH task for this

Tommy1231232: Good point @daywiss I'll make a separate CH ticket for this

daywiss (approved):

PR #21, opened 2021-05-17
Feat: Various text fixes

Tulun:
following have been addressed:

https://app.clubhouse.io/uma-project/story/832/denote-countdown-timer-in-hours-h-and-minutes-m

https://app.clubhouse.io/uma-project/story/833/when-you-click-on-the-settings-wheel-and-your-two-wallet-addresses-comes-up-call-the-cold-wallet-the-designatedvoting-wallet

https://app.clubhouse.io/uma-project/story/823/ui-issues-when-rejecting-transactions

https://app.clubhouse.io/uma-project/story/836/when-wallet-is-not-connected-show-instead-of-0-for-all-the-balances-in-wallet-module

https://app.clubhouse.io/uma-project/story/837/when-wallet-is-not-connected-show-display-for-the-vote-status

https://app.clubhouse.io/uma-project/story/844/style-alignment-issue-on-reveal

https://app.clubhouse.io/uma-project/story/846/formatting-issue-on-modal

https://app.clubhouse.io/uma-project/story/822/showing-2key-address

https://app.clubhouse.io/uma-project/story/872/add-spaces-between-usd-and-amount-in-wallet-module

https://app.clubhouse.io/uma-project/story/782/user-flow-for-not-signing-authorization-message

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { useContext, useEffect } from "react";
+import { useContext, useEffect, useState } from "react";
 import Router from "features/router";
 import { QueryClient } from "react-query";
 import usePrevious from "common/hooks/usePrevious";
@@ -8,13 +8,52 @@ import { ToastContainer, toast } from "react-toastify";
 import { ErrorContext } from "common/context/ErrorContext";
 import { OnboardContext } from "common/context/OnboardContext";
 
+import { recoverPublicKey } from "./features/vote/helpers/recoverPublicKey";
+import { derivePrivateKey } from "./features/vote/helpers/derivePrivateKey";
+
 interface Props {
   queryClient: QueryClient;
 }
 
+export interface SigningKeys {
+  [key: string]: {
+    publicKey: string;
+    privateKey: string;
+  };
+}
+
 function App(props: Props) {
+  const [signingKeys, setSigningKeys] = useState<SigningKeys>({});
+
   const { state, disconnect, dispatch } = useContext(OnboardContext);
   const { error, removeError } = useContext(ErrorContext);
+
+  useEffect(() => {
+    if (state.signer && state.address) {
+      const address = state.address;
+      const message = "Login to UMA Voter dApp";
+      const keyExists = signingKeys[address];
+      if (!keyExists) {
+        state.signer
+          .signMessage(message)
+          .then((msg) => {
+            const key = {} as { publicKey: string; privateKey: string };
+
+            const privateKey = derivePrivateKey(msg);
+            const publicKey = recoverPublicKey(privateKey);
+            key.privateKey = privateKey;
+            key.publicKey = publicKey;
+
+            setSigningKeys((prevKeys) => {
+              return { ...prevKeys, [address]: key };
+            });
+          })
+          .catch((err) => {
+            console.log("Sign failed");
+          });
+      }
+    }
+  }, [state.signer, state.address, signingKeys]);
 
   useEffect(() => {
     if (error)
@@ -40,7 +79,7 @@ function App(props: Props) {
 
   return (
     <div className="App">
-      <Router />
+      <Router signingKeys={signingKeys} />
       <ToastContainer
         position="top-right"
         autoClose={false}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,7 +26,7 @@ function App(props: Props) {
   const [signingKeys, setSigningKeys] = useState<SigningKeys>({});
 
   const { state, disconnect, dispatch } = useContext(OnboardContext);
-  const { error, removeError } = useContext(ErrorContext);
+  const { error, removeError, addError } = useContext(ErrorContext);
 
   useEffect(() => {
     if (state.signer && state.address) {
@@ -49,11 +49,11 @@ function App(props: Props) {
             });
           })
           .catch((err) => {
-            console.log("Sign failed");
+            addError("Sign failed.");
           });
       }
     }
-  }, [state.signer, state.address, signingKeys]);
+  }, [state.signer, state.address, signingKeys, addError]);
 
   useEffect(() => {
     if (error)
--- a/src/features/vote/CommitPhase.tsx
+++ b/src/features/vote/CommitPhase.tsx
@@ -68,7 +68,7 @@ const CommitPhase: FC<Props> = ({
   openViewDetailsModal,
 }) => {
   const [modalState, setModalState] = useState<SubmitModalState>("init");
-
+  const [submitErrorMessage, setSubmitErrorMessage] = useState("");
   const {
     state: { network, signer },
   } = useContext(OnboardContext);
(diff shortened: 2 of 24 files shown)


src/features/vote/PastViewDetailsModal.tsx
```diff
@@ -60,14 +60,19 @@ const _PastViewDetailsModal: ForwardRefRenderFunction<
   },
   externalRef
 ) => {
+  console.log("rewards claimed", rewardsClaimed);
```
daywiss, line 63: can prob remove

src/features/vote/useTableValues.ts
```diff
@@ -48,7 +51,29 @@ export default function useTableValues(
...
+          return {
+            ...el,
+            description,
+          };
+        })
+      );
+
+      descriptionsAdded.then((values) => {
+        setTableValues(values);
+      });
```
daywiss, line 76: do we need to catch error here?

src/features/vote/helpers/calculateTimeRemaining.ts
```diff
@@ -8,6 +8,12 @@ export const calculateTimeRemaining = () => {
   let text = "00:00";
   // format difference
   if (difference > 0) text = Duration.fromMillis(difference).toFormat("hh:mm");
+  // add h and m to time.
+  const split = text.split(":");
+  const ls = `${split[0]}h`;
+  const rs = `${split[1]}m`;
+
+  text = `${ls}:${rs}`;
```
daywiss, line 16: i think [this will work](https://moment.github.io/luxon/docs/manual/formatting.html#escaping)
`if (difference > 0) text = Duration.fromMillis(difference).toFormat("hh'h':mm'm'");`

src/features/vote/fetchUMIP.ts
```diff
@@ -0,0 +1,27 @@
...
+  number: number;
+};
+
+export const fetchUmip = async (number: number) => {
+  const ct = await contentfulClient.getEntries<UMIP>({
+    content_type: "umip",
+    "fields.number": number,
+  });
+  return ct.items[0].fields;
+};
```
daywiss, line 27: nice moving this out into just a regular ts file.  it would be nice to wrap this in a function or class so that code isnt just executed when the file is imported, but its probably not a big deal for now.

src/features/vote/useTableValues.ts
```diff
@@ -48,7 +51,29 @@ export default function useTableValues(
...
+        tv.map(async (el) => {
+          const isUmip = el.identifier.includes("Admin");
+          const umipNumber = isUmip
+            ? parseInt(el.identifier.split(" ")[1])
+            : undefined;
+
+          let description = "Price request.";
+
+          if (umipNumber) {
+            description = await (await fetchUmip(umipNumber)).description;
```
daywiss, line 64: if any of these fail, all of them will. might be ok, but consider handling the error somewhere if that is a problem

src/features/vote/useTableValues.ts
```diff
@@ -125,7 +150,31 @@ export default function useTableValues(
...
```
daywiss, line 164: same thing here, a single call could fail which invalidates all successful calls. you could also do something like [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled), which slightly changes what is returned, but at least you can get the results back and filter out errors.

src/features/vote/useTableValues.ts
```diff
@@ -125,7 +150,31 @@ export default function useTableValues(
...
+          return {
+            ...el,
+            description,
+          };
+        })
+      );
+
+      descriptionsAdded.then((values) => {
+        setTableValues(values);
+      });
```
daywiss, line 176: depending on what you do u may or may not need to catch error here

src/features/wallet/Wallet.tsx
```diff
@@ -67,7 +67,13 @@ const Wallet: FC<Props> = () => {
...
     network
   );
-  // const previousVotingAddress = usePrevious(votingAddress);
+
+  const signingPK =
+    hotAddress && signingKeys[hotAddress]
+      ? signingKeys[hotAddress].privateKey
+      : votingAddress && signingKeys[votingAddress]
+      ? signingKeys[votingAddress].privateKey
+      : "";
```
daywiss, line 76: i feel like this could be simplified somewhat:
`const signingPK = signingKeys[hotAddress]?.privateKey || signingKeys[votingAddress]?.privateKey || ''`
its ok to index into `signingKeys` with a undefined value, it should return undefined since you never set that, and you can use the optional chaining to prevent runtime error indexing into undefined

src/features/wallet/Wallet.tsx
```diff
@@ -163,10 +169,24 @@ const Wallet: FC<Props> = () => {
...
+                  {hotAddress ? "DVC: " : null} {shortenAddress(votingAddress)}
+                </VotingAddress>
+                {hotAddress ? (
+                  <VotingAddress>
+                    Voting: {shortenAddress(hotAddress)}
+                  </VotingAddress>
+                ) : null}
+                {signingPK ? (
+                  <Connected>
+                    Connected with {onboard?.getState().wallet.name}{" "}
```
daywiss, line 182: is `wallet.name` always going to exist? i feel like this could cause a crash, maybe safer to use optional chaining? wdyt

src/web3/post/commitVotes.ts
```diff
@@ -17,9 +17,10 @@ export const commitVotes = async (
     const tx = await contract.functions[
       "batchCommit((bytes32,uint256,bytes,bytes32,bytes)[])"
     ](data);
-    // console.log("commit votes TX?", tx);
+
     return tx;
   } catch (err) {
     console.log("Err in attempted commitVote transaction", err);
+    throw new Error(err.message);
```
daywiss, line 24: fyi you can just rethrow the old error: `throw err`. also do you want to keep the log here?

daywiss (commented):
looks good, mainly just have nits around error handling. some nice additions to modularize code

src/features/vote/CommitPhase.tsx
```diff
@@ -132,18 +132,29 @@ const CommitPhase: FC<Props> = ({
...
+            commitVotes(vc, fd)
+              .then((tx) => {
+                setModalState("pending");
+                setSubmitErrorMessage("");
+                // Need to confirm if the user submits the vote.
+                if (tx) {
+                  tx.wait(1)
+                    .then((conf: any) => {
+                      // Temporary, as mining is instant on local ganache.
+                      setTimeout(() => setModalState("success"), 5000);
```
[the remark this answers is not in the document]
Tulun: It's nice on localhost, but yeah, unnecessary probably for main / kovan.

Gamaranto (approved):
LGTM! All of the issues mentioned are addressed. One nit is that for the last issue, the circle stays gray and doesn't turn yellow as mentioned: https://app.clubhouse.io/uma-project/story/782/user-flow-for-not-signing-authorization-message

src/features/vote/useTableValues.ts
Tulun: I'll look into it, I think the last time I tried that it was a headache with TS.

src/features/wallet/Wallet.tsx
Tulun: I'm not sure -- Onboard's docs are not fully clear on this.

Tulun: We could have a fallback name if you'd like.

src/features/wallet/Wallet.tsx
Tulun: Null can't be used as a key, sadly. I could try to refactor so hotAddress is not string | null, though.

daywiss: ok i had a feeling that might have been the case. i think its fine as is then, dont bother changing

src/features/wallet/Wallet.tsx
daywiss: its les about a fallback name and more about indexing into an undefined.  id probably just want to check `getState()?.wallet?.name`

src/App.tsx
```diff
@@ -8,14 +8,53 @@ import { ToastContainer, toast } from "react-toastify";
...
   const { state, disconnect, dispatch } = useContext(OnboardContext);
   const { error, removeError } = useContext(ErrorContext);
 
+  useEffect(() => {
+    if (state.signer && state.address) {
+      const address = state.address;
+      const message = "Login to UMA Voter dApp";
+      const keyExists = signingKeys[address];
+      if (!keyExists) {
+        state.signer
```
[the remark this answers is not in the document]
Tulun: You can't use returns like this in useEffect. You have to return a cleanup function.

src/features/vote/fetchUMIP.ts
Tulun: I have built apiService type logic before as an object -- could always repeat.

src/features/vote/useTableValues.ts
Tulun: Yeah I want to test this more with the CMS -- should probably be done on Kovan if we can figure out how to organize the data.

src/App.tsx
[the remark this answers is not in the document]
daywiss: oh interesting, ok

src/features/vote/fetchUMIP.ts
daywiss: up to you, not critical at this point

daywiss (approved):
lgtm

PR #22, opened 2021-05-19
feat: events propagate errors to context; direct reward retrieve for single round;  total rewards can be read from old contracts

Tulun:
**Motivation**
1) Errors now propagate up to a notification for the user from event calls.
2) Total rewards can be read from past contracts on mainnet.
3) To save on gas, we avoid multicall.
4) Refresh app on network change to avoid any general app errors.

**Summary**

https://app.clubhouse.io/uma-project/story/835/rewards-are-not-accurate-ability-to-query-rewards-earned-from-past-contracts

https://app.clubhouse.io/uma-project/story/842/if-the-decryption-fails-the-dapp-does-not-propagate-that-error-to-the-user

https://app.clubhouse.io/uma-project/story/860/troubleshoot-why-new-voter-dapp-is-more-expensive-than-the-old-voter-dapp

https://app.clubhouse.io/uma-project/story/893/refresh-voter-dapp-on-network-change

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -269,17 +269,15 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                 <span
                   onClick={() => {
                     if (votingContract && votesRevealed && multicallContract) {
-                      const unclaimedRewards = votesRevealed.filter(function (
-                        obj
-                      ) {
-                        return !rewardsEvents.some(function (obj2) {
-                          return (
-                            obj.identifier === obj2.identifier &&
-                            obj.roundId === obj2.roundId &&
-                            obj.ancillaryData === obj2.ancillaryData
-                          );
-                        });
-                      });
+                      const unclaimedRewards = votesRevealed.filter(
+                        (x) =>
+                          !rewardsEvents.some(
+                            (y) =>
+                              x.identifier === y.identifier &&
+                              x.roundId === y.roundId &&
+                              x.ancillaryData === y.ancillaryData
+                          )
+                      );
 
                       collectRewards(
                         votingContract,
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -60,6 +60,7 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
     signer,
     address,
     network,
+    notify,
   } = useOnboard();
 
   const { addError } = useContext(ErrorContext);
@@ -281,16 +282,25 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                           )
                       );
 
-                      return collectRewards(
-                        votingContract,
-                        unclaimedRewards,
-                        multicallContract
-                      ) // wait for at least 1 block conf.
-                        .then((tx) => tx.wait(1))
-                        .then((conf: any) => {
-                          setAvailableRewards(DEFAULT_BALANCE);
-                        })
-                        .catch((err) => addError(err));
+                      return (
+                        collectRewards(
+                          votingContract,
+                          unclaimedRewards,
+                          multicallContract
+                        )
+                          // wait for at least 1 block conf.
+                          .then((tx) => {
+                            if (tx) {
+                              if (notify) notify.hash(tx.hash);
+
+                              tx.wait(1)
+                                .then((conf: any) => {
+                                  setAvailableRewards(DEFAULT_BALANCE);
+                                })
+                                .catch((err: any) => addError(err));
+                            }
+                          })
+                      );
                     }
                   }}
                   className="Wallet-collect"
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -284,9 +284,13 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                       return collectRewards(
                         votingContract,
                         unclaimedRewards,
-                        setAvailableRewards,
                         multicallContract
-                      ).catch((err) => addError(err));
+                      ) // wait for at least 1 block conf.
+                        .then((tx) => tx.wait(1))
+                        .then((conf: any) => {
+                          setAvailableRewards(DEFAULT_BALANCE);
+                        })
+                        .catch((err) => addError(err));
                     }
                   }}
                   className="Wallet-collect"
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -27,7 +27,8 @@ import calculateUMATotalValue from "./helpers/calculateUMATotalValue";
 import checkAvailableRewards from "./helpers/checkAvailableRewards";
 import collectRewards from "./helpers/collectRewards";
 import shortenAddress from "./helpers/shortenAddress";
-
+import { RewardsRetrieved } from "web3/get/queryRewardsRetrievedEvents";
+import { VoteRevealed } from "web3/get/queryVotesRevealedEvents";
 import {
   Wrapper,
   Connected,
@@ -87,14 +88,12 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
     hotAddress
   );
 
-  const { data: rewardsEvents } = useRewardsRetrievedEvents(
-    votingContract,
-    votingAddress
-  );
+  const { data: rewardsEvents = [] as RewardsRetrieved[] } =
+    useRewardsRetrievedEvents(votingContract, votingAddress);
 
   const { multicallContract } = useMulticall(signer, isConnected, network);
 
-  const { data: votesRevealed } = useVotesRevealedEvents(
+  const { data: votesRevealed = [] as VoteRevealed[] } = useVotesRevealedEvents(
     votingContract,
     votingAddress
   );
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -185,7 +185,7 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                 ) : (
                   <Reconnect>
                     {" "}
-                    Sign in rejected. Reconnect to sign in.
+                    Account needs to sign message. Reconnect if sign cancelled.
                   </Reconnect>
                 )}
               </>
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -287,9 +287,7 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                         unclaimedRewards,
                         setAvailableRewards,
                         multicallContract
-                      ).catch((err) => {
-                        addError(err.message);
-                      });
+                      ).catch((err) => addError(err));
                     }
                   }}
                   className="Wallet-collect"
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -269,9 +269,21 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                 <span
                   onClick={() => {
                     if (votingContract && votesRevealed && multicallContract) {
+                      const unclaimedRewards = votesRevealed.filter(function (
+                        obj
+                      ) {
+                        return !rewardsEvents.some(function (obj2) {
+                          return (
+                            obj.identifier === obj2.identifier &&
+                            obj.roundId === obj2.roundId &&
+                            obj.ancillaryData === obj2.ancillaryData
+                          );
+                        });
+                      });
+
                       collectRewards(
                         votingContract,
-                        votesRevealed,
+                        unclaimedRewards,
                         setAvailableRewards,
                         multicallContract
                       );
--- a/src/features/wallet/Wallet.tsx
+++ b/src/features/wallet/Wallet.tsx
@@ -1,5 +1,5 @@
 /** @jsxImportSource @emotion/react */
-import { FC, useState, useEffect } from "react";
+import { FC, useState, useEffect, useContext } from "react";
 import tw from "twin.macro"; // eslint-disable-line
 import { ethers } from "ethers";
 import useModal from "common/hooks/useModal";
@@ -19,6 +19,7 @@ import {
 } from "hooks";
 
 import TwoKeyContractModal from "./TwoKeyContractModal";
+import { ErrorContext } from "common/context/ErrorContext";
 
 // Helpers
 import formatWalletBalance from "./helpers/formatWalletBalance";
@@ -59,6 +60,8 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
     address,
     network,
   } = useOnboard();
+
+  const { addError } = useContext(ErrorContext);
 
   const prevTotalCollected = usePrevious(totalUmaCollected);
   const prevAvailableReweards = usePrevious(availableRewards);
@@ -279,12 +282,14 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                           )
                       );
 
-                      collectRewards(
+                      return collectRewards(
                         votingContract,
                         unclaimedRewards,
                         setAvailableRewards,
                         multicallContract
-                      );
+                      ).catch((err) => {
+                        addError(err.message);
+                      });
                     }
                   }}
                   className="Wallet-collect"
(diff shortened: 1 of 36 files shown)


src/features/wallet/helpers/collectRewards.ts
```diff
@@ -32,46 +33,80 @@ export default function collectRewards(
...
-    )
-      // wait for at least 1 block conf.
+    return retrieveRewards(contract, postData)
       .then((tx) => tx.wait(1))
       .then((conf: any) => {
         setAvailableRewards(DEFAULT_BALANCE);
       })
-      .catch((err) => console.log("err in retrieve rewards", err))
-  );
+      .catch((err) => console.log("err in retrieve rewards", err));
```
daywiss, line 110: could you bubble this error up to the user?

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -32,46 +33,80 @@ export default function collectRewards(
...
+    return (
+      multicallContract.functions["aggregate((address,bytes)[])"](
+        multicallCollectRewards
+      )
+        // wait for at least 1 block conf.
+        .then((tx) => tx.wait(1))
+        .then((conf: any) => {
+          setAvailableRewards(DEFAULT_BALANCE);
+        })
+        .catch((err) => console.log("err in retrieve rewards", err))
```
daywiss, line 79: could we bubble this up to user?

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -32,46 +33,80 @@ export default function collectRewards(
...
-  // console.log("multi calls sorted?", multicallCollectRewards);
+      postData.pendingRequests = pendingRequestData;
+    });
 
-  return (
-    multicallContract.functions["aggregate((address,bytes)[])"](
-      multicallCollectRewards
-    )
-      // wait for at least 1 block conf.
+    return retrieveRewards(contract, postData)
```
daywiss, line 105: would be nice if we could functionalize these further so `collectRewards` could call `collectManyRewards` and `collectOneReward` which both return a promise where you can de-duplicate promise handling logic:
```js
let collect // typing this might be difficult
if (uniqueRoundIds.length > 1) {
  collect = collectManyRewards(data, uniqueRoundIds,pendingRequests, ...etc)
}else{
  collect = collectOneRewards(data, uniqueRoundIds,pendingRequests, ...etc)
}
collect.then((tx) => tx.wait(1))
      .then((conf: any) => {
        setAvailableRewards(DEFAULT_BALANCE);
      })
      .catch((err) => console.log("err in retrieve rewards", err));
```

src/hooks/useCurrentRoundId.ts
```diff
@@ -12,16 +14,23 @@ const contract = createVoidSignerVotingContractInstance(
...
+      return queryCurrentRoundId(contract)
+        .then((res) => {
+          if (res) {
+            return res;
+          } else {
+            return "";
+          }
+        })
+        .catch((err) => {
+          addError(err.message);
```
daywiss, line 31: would be great if `addError` could take the full error object and pull the message out in the context. this way you could cleanly call `.catch(addError)` in some cases (not necessarily in this case)

src/hooks/useCurrentRoundId.ts
```diff
@@ -12,16 +14,23 @@ const contract = createVoidSignerVotingContractInstance(
...
           return "";
-        }
-      });
+        });
```
daywiss, line 33: i've mentioned this a while ago, but this general react-query pattern is used a lot in this code, but i really think it could be simplified if you require the view to always check if data exists before using it.  Most of this file could be squashed into a few lines:
```js
export default function useCurrentRoundId() {
  const { addError } = useContext(ErrorContext);

  return useQuery<string>(
    "roundId",
    () => {
      return queryCurrentRoundId(contract).catch(addError)  // addError is optional here, you could let this throw to pass error to view in some situations
    }
  );
}
```

The only changes are `addError` accepts a full error object rather than just a string and any view that consumes this hook must check that `data` exists before using it. what do you think about this?

src/hooks/usePendingRequests.ts
```diff
@@ -26,23 +26,24 @@ export default function usePendingRequests() {
...
+          if (res) {
+            return res;
+          } else {
+            return [];
+          }
+        })
+        .catch((err) => {
+          addError(error!.message);
+          return [] as PendingRequest[];
+        });
```
daywiss, line 40: same comment as above, this could be simplified if we assume data can be undefined in the view, and its checked there. you could even have the view be responsible for setting the defaults: `{data=[]} = usePendingRequests()`

src/web3/get/queryRewardsRetrievedEvents.ts
```diff
@@ -56,6 +59,53 @@ export const queryRewardsRetrieved = async (
...
     });
+
+    // On mainnet, look back at old voting contracts to retrieve rewards
+    // for voters who have voted on previous iteration of the contract.
+    // this is largely just for summation of Total UMA Collected.
+    if (
+      process.env.REACT_APP_CURRENT_ENV === "main" ||
+      process.env.REACT_APP_CURRENT_ENV === undefined
+    ) {
+      const promises = OLD_VOTING_CONTRACT_ADDRESSES.map(async (addr) => {
```
daywiss, line 70: can we generalize this to a function which takes any voting contract address? Or at least a function which handles current voting contract and previous ones.

`async getRewards(userAddress,contractAddress)`

and then do something like this
```js
const rewards = await Promise.all(
  allVotingContractAddresses.map(contractAddress=>getRewards(userAddress,contractAddress,provider))
)
return rewards.flat()
```

wdyt?

src/web3/get/queryRewardsRetrievedEvents.ts
```diff
@@ -56,6 +59,53 @@ export const queryRewardsRetrieved = async (
...
+      });
+
+      return Promise.all(promises).then((results) => {
+        const flatten = results.flat();
+        return [...rewards, ...flatten];
+      });
+    }
+    return rewards;
   } catch (err) {
     console.log("err", err);
```
daywiss, line 110: can we bubble this up

src/web3/post/revealVotes.ts
```diff
@@ -23,7 +23,7 @@ export const revealVotes = async (
     const tx = await contract.functions[
       "batchReveal((bytes32,uint256,int256,bytes,int256)[])"
     ](data);
-    console.log("successfully Revealed", tx);
+
     return tx;
   } catch (err) {
     console.log("err in Reveal votes", err);
```
daywiss, line 29: can we bubble this to user?

daywiss (commented):
looks good, i have some ideas on generalizing some of the logic, want to get your thoughts

src/features/wallet/Wallet.tsx
```diff
@@ -269,9 +269,19 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
...
                     if (votingContract && votesRevealed && multicallContract) {
+                      const unclaimedRewards = votesRevealed.filter(
+                        (x) =>
+                          !rewardsEvents.some(
+                            (y) =>
+                              x.identifier === y.identifier &&
+                              x.roundId === y.roundId &&
+                              x.ancillaryData === y.ancillaryData
+                          )
+                      );
```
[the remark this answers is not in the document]
Tulun: We could store it in a local variable, but it's only really needed to be calculated when we actually submit the transaction.

src/features/wallet/helpers/collectRewards.ts
Tulun: Yup.

src/hooks/useCurrentRoundId.ts
Tulun: I guess if we wanted to add custom messages (in some weird case where the error is nonsensical without context) I guess we could just throw a new error object we construct.

src/hooks/useCurrentRoundId.ts
Tulun: Would you suggest I pass in empty arrays in places it's looking for the data? Like if it's voteData, it's actually voteData || [] ?

src/web3/get/queryRewardsRetrievedEvents.ts
Tulun: I think the issue is the ABI is different between the newest contract and oldest. That's why I'm forced to create a new version of the old contract.

i could add the ABI in an arg, but it gets messier somewhat quickly. Needing to reference the old contract(s) is awkward.

src/web3/get/queryRewardsRetrievedEvents.ts
Tulun: Done!

src/web3/post/revealVotes.ts
Tulun: Done!

src/hooks/usePendingRequests.ts
Tulun: +1

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -32,46 +33,84 @@ export default function collectRewards(
...
+          pendingRequest.ancillaryData = datum.ancillaryData
+            ? datum.ancillaryData
+            : "0x";
+          pendingRequest.identifier = ethers.utils.toUtf8Bytes(
+            datum.identifier
+          );
+          pendingRequest.time = datum.time;
+          pendingRequestData.push(pendingRequest);
+        }
+      });
```
mrice32, line 102: nit: seems like this might be a bit simpler -- I think the looping structure isn't as necessary here as it is with the multicall:
```ts
    const [roundId] = uniqueRoundIds;
    const postData.pendingRequests = data.filter((datum) => datum.roundId === roundId)
      .map((datum): PendingRequestRetrieveReward  => ({
        ...datum,
        ancillaryData: datum.ancillaryData ? datum.ancillaryData : "0x",
        identifier: ethers.utils.toUtf8Bytes(datum.identifier)
      })
    );

File: src/features/wallet/helpers/collectRewards.ts [hunk 12]
```diff
@@ -32,46 +33,84 @@ export default function collectRewards(
...
-    const pendingRequestData = [] as PendingRequestRetrieveReward[];
+  // Do a multicall request if the user is collecting from multiple rounds.
+  // Otherwise just call the rewards function directly.
+  if (uniqueRoundIds.length > 1) {
+    uniqueRoundIds.forEach((roundId) => {
+      const mcr = {} as MulticallCollectRewards;
+      const postData = {} as PostRetrieveReward;
+      const pendingRequestData = [] as PendingRequestRetrieveReward[];
+
+      data.forEach((datum) => {
```
mrice32, line 44: Similar to below, I think this might be a little simpler with a filter + map so you don't have to initialize an array then append to it.

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -32,46 +33,84 @@ export default function collectRewards(
...
+        .then((tx) => tx.wait(1))
+        .then((conf: any) => {
+          setAvailableRewards(DEFAULT_BALANCE);
+        })
+        .catch((err) => {
+          throw err;
+        })
     );
+  } else {
+    const postData = {} as PostRetrieveReward;
```
mrice32, line 84: Sorry, maybe I'm missing something, but where are we setting the address and roundId in this postData struct?

src/features/wallet/helpers/collectRewards.ts
mrice32: If you don't love the chained calls (they aren't super readable), you could do something more like:
```ts
const filteredRequests = data.filter((datum) => datum.roundId === roundId);
postData.pendingRequests = filteredRequests.map(...);
```

src/features/wallet/Wallet.tsx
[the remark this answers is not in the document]
Tulun: We show what rewards they have available to get yes. But that's done by static call.

src/features/wallet/helpers/collectRewards.ts
Tulun: Wouldn't it be simpler to return those functions?

Tulun: Maybe we should move the set visual state into the wallet file anyway.

src/features/wallet/helpers/collectRewards.ts
Tulun: Yup, that works great.

src/features/wallet/helpers/collectRewards.ts
Tulun: Yep.

src/features/wallet/helpers/collectRewards.ts
Tulun: Refactored and works great.

src/features/wallet/helpers/collectRewards.ts
mrice32: Can you point me to where?

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -20,38 +18,79 @@ interface MulticallCollectRewards {
...
+  data: RewardsRetrieved[],
+  // Type returned by useState variable in Wallet.tsx,
+  roundId: string,
+  contract: ethers.Contract
+) {
+  const postData = {
+    voterAddress: data[0].address,
+    roundId,
+  } as PostRetrieveReward;
+
```
Tulun, line 53: @mrice32

src/features/wallet/helpers/collectRewards.ts
```diff
@@ -20,38 +18,79 @@ interface MulticallCollectRewards {
...
+  const votingInterface = new ethers.utils.Interface(VotingArtifact.abi);
+
   uniqueRoundIds.forEach((roundId) => {
     const mcr = {} as MulticallCollectRewards;
-    const postData = {} as PostRetrieveReward;
-    const pendingRequestData = [] as PendingRequestRetrieveReward[];
+    const postData = {
+      voterAddress: data[0].address,
+      roundId,
+    } as PostRetrieveReward;
```
Tulun, line 82: @mrice32

src/web3/get/queryEncryptedVotesEvents.ts
```diff
@@ -95,6 +95,6 @@ export const queryEncryptedVotes = async (
     );
     return decryptedEvents;
   } catch (err) {
-    console.log("err", err);
+    throw err;
```
daywiss, line 98: why do we catch this block of code only to rethrow? just remove try catch

src/web3/get/queryGetPendingRequests.ts
```diff
@@ -47,7 +46,6 @@ export const queryGetPendingRequests = async (contract: ethers.Contract) => {
       return [] as PendingRequest[];
     }
   } catch (err) {
-    console.log("err", err);
-    throw new Error(err.message);
+    throw err;
```
daywiss, line 49: same for this block

src/web3/post/revealVotes.ts
```diff
@@ -23,9 +23,9 @@ export const revealVotes = async (
     const tx = await contract.functions[
       "batchReveal((bytes32,uint256,int256,bytes,int256)[])"
     ](data);
-    console.log("successfully Revealed", tx);
+
     return tx;
   } catch (err) {
-    console.log("err in Reveal votes", err);
+    throw err;
```
daywiss, line 29: can also remove try catch here if we arent doing anything with error

daywiss (approved):
thanks for doing this extensive refactor! i think its a good improvement. i only have small nits, but use your judgement on them

src/features/wallet/Wallet.tsx
```diff
@@ -269,12 +271,26 @@ const Wallet: FC<Props> = ({ signingKeys }) => {
                 <span
                   onClick={() => {
                     if (votingContract && votesRevealed && multicallContract) {
-                      collectRewards(
+                      const unclaimedRewards = votesRevealed.filter(
```
mrice32, line 274: Does this fix the multicall gas issue?

mrice32 (commented):
Looks good as long as @daywiss's comments are resolved!!

src/features/wallet/Wallet.tsx
Tulun: I think before it wasn't filtering between what you had claimed and what was revealed. 

So now say you have 40 claimed events, but 42 reveals -- obviously those 2 reveals are the eligible claimed rewards to try to claim

PR #31, opened 2021-06-01
feat: Adjust commit modal flow

Tulun:
1) Commit modal flow closes and leverage Notify.js only.

2) Removed reference to modal state.

### DIFF
--- a/src/features/vote/CommitPhase.tsx
+++ b/src/features/vote/CommitPhase.tsx
@@ -143,7 +143,6 @@ const CommitPhase: FC<Props> = ({
           if (vc) {
             commitVotes(vc, fd)
               .then((tx) => {
-                // console.log("tx", tx);
                 setSubmitErrorMessage("");
                 close();
                 reset();
@@ -158,7 +157,7 @@ const CommitPhase: FC<Props> = ({
                     reset();
                   })
                   .catch((err: any) => {
-                    setSubmitErrorMessage("Error with tx.");
+                    setSubmitErrorMessage(`Error with tx: ${err.message}`);
                   });
               })
               .catch((err) => {
--- a/src/features/vote/CommitPhase.tsx
+++ b/src/features/vote/CommitPhase.tsx
@@ -145,8 +145,9 @@ const CommitPhase: FC<Props> = ({
             commitVotes(vc, fd)
               .then((tx) => {
                 // console.log("tx", tx);
-                setModalState("pending");
+                // setModalState("pending");
                 setSubmitErrorMessage("");
+                close();
 
                 // Need to confirm if the user submits the vote.
                 assert(tx, "Transaction did not get submitted, try again");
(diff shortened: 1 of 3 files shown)


daywiss:
would be great to know what the original issue that motivated this and then roughly what was changed to fix it

src/features/vote/CommitPhase.tsx
```diff
@@ -145,30 +144,25 @@ const CommitPhase: FC<Props> = ({
             commitVotes(vc, fd)
               .then((tx) => {
                 // console.log("tx", tx);
```
daywiss, line 146: could probably clean this up

src/features/vote/CommitPhase.tsx
```diff
@@ -145,30 +144,25 @@ const CommitPhase: FC<Props> = ({
...
-                setModalState("pending");
                 setSubmitErrorMessage("");
-
+                close();
+                reset();
```
daywiss, line 149: so my understanding is instead of leaving modal open after commit, it gets closed and notify js will handle the rest?

src/features/vote/CommitPhase.tsx
```diff
@@ -145,30 +144,25 @@ const CommitPhase: FC<Props> = ({
...
                 tx.wait(1)
                   .then((conf: any) => {
-                    // Temporary, as mining is instant on local ganache.
-                    // setTimeout(() => setModalState("success"), 5000);
-                    setModalState("success");
                     refetchEncryptedVotes();
                     reset();
                   })
                   .catch((err: any) => {
                     setSubmitErrorMessage("Error with tx.");
```
daywiss, line 161: small nit here, could we add the error message. would be useful if there actually is an error by a user to know what the underlying error was when reported: `setSubmitErrorMessage('Error with tx: ' + err.message);`

src/features/vote/CommitPhase.tsx
```diff
@@ -375,15 +368,11 @@ const CommitPhase: FC<Props> = ({
         isOpen={isOpen}
         close={close}
         ref={modalRef}
-        modalState={modalState}
-        setModalState={setModalState}
```
daywiss, line 379: how come this was no longer needed? just curious

daywiss (approved):
looks good, just some questions and a nits. looking forward to trying this out

src/features/vote/CommitPhase.tsx
Tulun: Yeah exactly.

src/features/vote/CommitPhase.tsx
Tulun: Sounds good!

src/features/vote/CommitPhase.tsx
Tulun: the modal state doesn't have the pending and success states, which were related to the transaction, which is now handled by Notify.

PR #34, opened 2021-06-03
Feat: signing keys stored in local storage

Tulun:
Implementation of keeping signing keys in local storage so user should never have to resign unless we change the message.

https://app.clubhouse.io/uma-project/story/863/never-or-almost-never-need-to-re-sign-when-using-the-same-browser-on-the-same-machine

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,7 +61,7 @@ function App(props: Props) {
 
               const updatedKeys = {
                 ...keys,
-                [hashedMessage]: { [address]: key },
+                [hashedMessage]: { ...keys[hashedMessage], [address]: key },
               };
 
               localStorage.setItem(
@@ -100,6 +100,8 @@ function App(props: Props) {
             addError(error);
           });
       }
+    } else {
+      setSigningKeys({});
     }
   }, [state.signer, state.address, addError, currentRoundId]);
 
@@ -128,6 +130,7 @@ function App(props: Props) {
       state.network &&
       state.network.chainId !== previousNetwork.chainId
     ) {
+      debugger;
       window.location.reload();
     }
   }, [state.network, previousNetwork]);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -130,7 +130,6 @@ function App(props: Props) {
       state.network &&
       state.network.chainId !== previousNetwork.chainId
     ) {
-      debugger;
       window.location.reload();
     }
   }, [state.network, previousNetwork]);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -81,8 +81,14 @@ function App(props: Props) {
     [addError]
   );
 
+  const previousRoundId = usePrevious(currentRoundId);
   useEffect(() => {
-    if (state.signer && state.address && currentRoundId && addError) {
+    if (
+      state.signer &&
+      state.address &&
+      currentRoundId &&
+      currentRoundId !== previousRoundId
+    ) {
       const address = state.address;
       const hashedMessage = currentSigningMessage(
         Number(currentRoundId)
@@ -96,17 +102,24 @@ function App(props: Props) {
         if (!keyExists) {
           signMessage(keys, state.address, state.signer, currentRoundId);
         } else {
-          console.log("here thrice");
           setSigningKeys(keys);
         }
       } catch (err) {
         const keys = {} as SigningKeys;
         signMessage(keys, state.address, state.signer, currentRoundId);
       }
-    } else {
+    }
+    if (!state.address || !state.signer) {
       setSigningKeys({});
     }
-  }, [state.signer, state.address, addError, currentRoundId, signMessage]);
+  }, [
+    state.signer,
+    state.address,
+    addError,
+    currentRoundId,
+    signMessage,
+    previousRoundId,
+  ]);
 
   useEffect(() => {
     if (error)
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useState } from "react";
+import { useContext, useEffect, useState, useCallback } from "react";
 import Router from "features/router";
 import { QueryClient } from "react-query";
 import usePrevious from "common/hooks/usePrevious";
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -11,29 +12,73 @@ import { OnboardContext } from "common/context/OnboardContext";
...
+  const { data: currentRoundId } = useCurrentRoundId();
 
   useEffect(() => {
     if (state.signer && state.address) {
       const address = state.address;
-      const message = "Login to UMA Voter dApp";
-      const keyExists = signingKeys[address];
-      if (!keyExists) {
+      const message = currentSigningMessage(Number(currentRoundId));
+      const hashedMessage = ethers.utils.formatBytes32String(message);
```
mrice32, line 46: Does formatBytes32String perform a hash? According to the docs, it doesn't seem to? It appears as if this is just taking a utf8 string and converting it to bytes32? 

Also, are we assuming the message won't be longer than 32 bytes?

src/App.tsx
```diff
@@ -11,29 +12,73 @@ import { OnboardContext } from "common/context/OnboardContext";
...
+      const keysString = localStorage.getItem(SIGNING_KEYS_STORAGE_KEY);
+      if (keysString) {
+        const keys = JSON.parse(keysString) as SigningKeys;
+        const keyExists = keys[hashedMessage][address];
```
mrice32, line 50: Thoughts on putting some try-catches around this logic? Seems like we'd want an escape hatch if anything fails so we always fall back to just re-requesting the signature. Thoughts? Or do you think that could cause issues?

Total side note: as a long-term security measure to avoid someone somehow corrupting local storage to break a voter's reveal (however unlikely that may be), we could add a step where we validate the private key -- we could do this similarly to how we verify signatures on-chain for the reveal snapshot. How likely do you think it is that someone is able to corrupt local storage? Could a scammy extension do this?

src/App.tsx
```diff
@@ -11,29 +12,73 @@ import { OnboardContext } from "common/context/OnboardContext";
...
+        if (!keyExists) {
+          state.signer
+            .signMessage(message)
+            .then((msg) => {
+              const key = {} as { publicKey: string; privateKey: string };
```
mrice32, line 55: Is this code block the same as below? Could we make this a separate function or somehow reuse this code so it's not repeated?

src/common/currentSigningMessage.ts
```diff
@@ -0,0 +1,14 @@
+export default function currentSigningMessage(currentRound: number) {
```
mrice32, line 1: +1 -- very nice. This will make it super easy to quickly modify the message with no fallout.

src/features/vote/ActiveRequests.tsx
```diff
@@ -84,11 +86,18 @@ const ActiveRequests: FC<Props> = ({
     return () => clearInterval(timer);
   }, []);
 
+  const message = currentSigningMessage(Number(roundId));
+  const hashedMessage = ethers.utils.formatBytes32String(message);
```
mrice32, line 90: Same question about the hash as above. Also, should we have `currentSigningMessage` return the hashed message if that's the only one we use in the logic?

mrice32 (commented):
Looks great! A few very minor comments.

src/App.tsx
Tulun: Fair, I can use a web3 func instead.

src/App.tsx
Tulun: Yeah the issue with your security hole is a problem. We are only looking if a signing key exists -- if someone got access to the localStorage they could scramble the value. The only fix would be to re-sign, which they could by clearing localstorage.

now the real question is, progmatically, I'm not sure how we know their signing key is bad. We could see if the MAC reveal fails on that particular key then ask them to re-sign in the decryption phase?

mrice32: There's a more comprehensive way to do this, but it's a bit involved.

You can do an operation called a "recover" where if you know the data that was signed (we do) you can back out the original public key (or address) that generated that signature (if it's a valid signature at all).

See this web3 method: https://web3js.readthedocs.io/en/v1.2.11/web3-eth-personal.html#ecrecover. Note: like our other crypto work, this might be tricky to get right, but technically it should be the inverse of this line: https://github.com/UMAprotocol/protocol/blob/4e98402896fe50a5013c8decfaa0f261363ae33c/packages/common/src/Crypto.js#L39. However, because we only need the signature and message, we don't need to ask the user for anythin to do the verification.

Once again, this is something we should add later, since it's not likely in the short term and might be a little tricky to get right.

daywiss (changes requested):

daywiss (approved):
thanks, looks good, nice improvement

src/common/currentSigningMessage.ts
```diff
@@ -0,0 +1,25 @@
...
+    return {
+      message: SIGNING_MESSAGE_ONE,
+      hashedMessage: web3.utils.utf8ToHex(SIGNING_MESSAGE_ONE),
+    };
+  }
+
+  // Default return
+  return {
+    message: SIGNING_MESSAGE_ONE,
+    hashedMessage: web3.utils.utf8ToHex(SIGNING_MESSAGE_ONE),
```
mrice32, line 20: Similar question as before -- I don't think this is a hash? Not saying we even need to hash anything! I just don't think we should call this a hash.

For reference, something like `web3.utils.sha3` or `web3.utils.soliditySha3` are hash functions: https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#sha3.

mrice32 (approved):
Only one minor comment about the hash naming. LGTM once that's figured out!

Gamaranto:
What is the status of this? @Tulun @mrice32 is this something we still want to add or should we go ahead and close this?

mrice32:
> What is the status of this? @Tulun @mrice32 is this something we still want to add or should we go ahead and close this?

I think we should add this to prod as soon as @Tulun is comfortable! Would be a really nice feature.

Tulun (commented):
> > What is the status of this? @Tulun @mrice32 is this something we still want to add or should we go ahead and close this?
> 
> I think we should add this to prod as soon as @Tulun is comfortable! Would be a really nice feature.

We need to rethink this a bit, as we currently also want the new app to be backwards compatible with the old app.

mrice32:
> > > What is the status of this? @Tulun @mrice32 is this something we still want to add or should we go ahead and close this?
> > 
> > 
> > I think we should add this to prod as soon as @Tulun is comfortable! Would be a really nice feature.
> 
> We need to rethink this a bit, as we currently also want the new app to be backwards compatible with the old app.

@Tulun Isn't the caching of the signature separate from the message generation, though? So if the message changes each round, you at least only have to sign once per round? And then we could move to a new message later?

PR #35, opened 2021-06-04
bug: hotfix for regexp

Tulun:
Regexp was too strict, need to allow Admin votes.


daywiss (approved):
tested this on current vote, seems to work thanks

PR #45, opened 2021-06-14
feat: Sign message equivalent to old voter dapp. Added voting address to modal.

Tulun:
Features as discussed in earlier morning meeting as requirements for public app release.


daywiss (approved):
lgtm, only thing ill mention is that it would be nice to show total uma voted with on confirmation page, but this is not strictly necessary. if needed we can always add later @Tommy1231232

PR #47, opened 2021-06-17
Feat: Rewards wallet fixes

Tulun:
Following fixed:

1. Adjusted connected circle to be more centred and changed colour. ( https://app.clubhouse.io/uma-project/story/1387/center-status-button-for-wallet)
Screen Shot 2021-06-17 at 4 05 17 PM
https://app.clubhouse.io/uma-project/story/1379/update-color-of-the-connected-with-metamask-indicator-to-green

2. More strict cond on the refetching of this data and clear cache on account change. (https://app.clubhouse.io/uma-project/story/1304/receiving-bad-mac-when-switching-between-wallets-in-top-right-corner)

3. Make a refetch of past requests when wallet balances change. (https://app.clubhouse.io/uma-project/story/1309/rewards-from-last-week-not-appearing-anywhere-in-ui)

### DIFF
--- a/src/features/router/Router.tsx
+++ b/src/features/router/Router.tsx
@@ -11,21 +11,31 @@ import Wallet from "features/wallet";
 import Footer from "common/components/footer";
 import { SigningKeys } from "App";
 
+import useVoteData, { PriceRequestRound } from "common/hooks/useVoteData";
+
 interface Props {
   signingKeys: SigningKeys;
 }
 
 const Router: FC<Props> = ({ signingKeys }) => {
+  const {
+    data: voteSummaryData = [] as PriceRequestRound[],
+    refetch: refetchVoteSummaryData,
+  } = useVoteData();
+
   return (
     <BrowserRouter>
       <div>
         <Navbar />
         {/* A <Switch> looks through its children <Route>s and
             renders the first one that matches the current URL. */}
-        <Wallet signingKeys={signingKeys} />
+        <Wallet
+          signingKeys={signingKeys}
+          refetchVoteSummaryData={refetchVoteSummaryData}
+        />
         <Switch>
           <Route path="/">
-            <Vote signingKeys={signingKeys} />
+            <Vote signingKeys={signingKeys} voteSummaryData={voteSummaryData} />
           </Route>
         </Switch>
       </div>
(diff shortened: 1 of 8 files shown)


src/features/router/Router.tsx
```diff
@@ -11,21 +11,31 @@ import Wallet from "features/wallet";
...
             renders the first one that matches the current URL. */}
-        <Wallet signingKeys={signingKeys} />
+        <Wallet
+          signingKeys={signingKeys}
+          refetchVoteSummaryData={refetchVoteSummaryData}
+        />
         <Switch>
           <Route path="/">
-            <Vote signingKeys={signingKeys} />
+            <Vote signingKeys={signingKeys} voteSummaryData={voteSummaryData} />
```
daywiss, line 38: just curious why this was moved to a prop

src/hooks/useEncryptedVotesEvents.ts
```diff
@@ -15,6 +16,12 @@ export default function useEncryptedVotesEvents(
 ) {
   const { addError } = useContext(ErrorContext);
 
+  const previousAddress = usePrevious(address);
+  // To prevent a "bad MAC" error, try not to run the query if the previousAddress is defined but it doesn't equal current address.
+  // This happens because the wallet state is async -- sometimes when disconnecting, address is defined before we change the state.
+  const addressCheck = !(
+    previousAddress !== null && previousAddress !== address
+  );
```
daywiss, line 24: im not sure i follow this logic, are you sure it works?  so it will query if:
1. previousAddress == null
2. previousAddress == address
for 1. its probably fine, but is it the case that 2 will ever be true, maybe i dont understand what previousAddress represents

src/features/router/Router.tsx
Tulun: Didn't want to recreate the hook in two locations, when I was just passing down a refetch.

src/hooks/useEncryptedVotesEvents.ts
Tulun: So sometimes we will have the address (only when. they've connected) but they've *changed* addresses.

This leads to the Bad MAC error. Basically, as other parts of the app disconnect them as they change address, I don't want this race condition to show them an error which shouldn't concern them.

daywiss (approved):

PR #48, opened 2021-06-21
feat: Voter stats in reveal and commit modal.

Tulun:
Voter summary stats added to commit and reveal modals: (https://app.clubhouse.io/uma-project/story/1278/add-vote-statistics)

Screen Shot 2021-06-21 at 2 17 40 PM

### DIFF
--- a/src/features/vote/ActiveViewDetailsModal.tsx
+++ b/src/features/vote/ActiveViewDetailsModal.tsx
@@ -19,6 +19,9 @@ import {
   IconsItem,
   Icon,
   LastStateValue,
+  StateValueAddress,
+  RevealHeader,
+  RevealPercentage,
 } from "./styled/DetailModals.styled";
 import { ModalState } from "./ActiveRequests";
 import { DiscordRed, CopyIcon } from "assets/icons";
@@ -29,6 +32,7 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import ReactTooltip from "react-tooltip";
 import web3 from "web3";
 import { VoteEvent } from "web3/types.web3";
+import { VoteRevealed } from "web3/get/queryVotesRevealedEvents";
 
 interface Props {
   isOpen: boolean;
@@ -40,9 +44,13 @@ interface Props {
   ancData: string;
   unix: string;
   committedVotes: VoteEvent[] | undefined | void;
+  revealedEvents: VoteRevealed[] | undefined | void;
+  roundId: string;
 }
 
 const NULL_ANC_DATA = "0x";
+const NULL_NUM_COMMITTED_VOTES = 0;
+const NULL_PERCENTAGE = "0.00";
 
 const _ActiveViewDetailsModal: ForwardRefRenderFunction<
   HTMLElement,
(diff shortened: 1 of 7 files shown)


src/features/vote/ActiveViewDetailsModal.tsx
```diff
@@ -38,24 +43,85 @@ interface Props {
...
   });
 
+  useEffect(() => {
+    if (committedVotes && committedVotes.length && roundId && proposal) {
+      const findCommitsForProposal = committedVotes.filter(
+        (x) =>
+          x.identifier === proposal && x.roundId === roundId && x.time === unix
+      );
+      const uniqueCommitsByAddress = findCommitsForProposal.filter(
+        (v, i, a) => a.findIndex((t) => t.address === v.address) === i
```
daywiss, line 92: i might need an explanation of this logic, its a bit obscure, but it looks like you are trying to only count addresses once, and ignore duplicates? If this is the case you might want to try a [js set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set). You can initialize the set with an array of all addresses, and then use `.size` to get the length of only uniques

src/features/vote/ActiveViewDetailsModal.tsx
```diff
@@ -38,24 +43,85 @@ interface Props {
...
+    }
+  }, [revealedEvents, roundId, proposal, unix]);
+
+  useEffect(() => {
+    if (numberCommittedVotes > 0) {
+      const percentage = (
+        (numberRevealedAddresses / numberCommittedVotes) *
+        100
+      ).toFixed(2);
+      setRevealPercentage(percentage);
```
daywiss, line 119: these stats like functions seem appropriate for a utility file that can be unit tested, though they might not be complex enough to warrant that. I could imagine a collection of static functions which can then be composed into a function that returns the end state you want, ie the final display variables. wdyt?

src/hooks/useVotesRevealedEventsSummary.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const contract = createVoidSignerVotingContractInstance(
+  provider,
+  determineBlockchainNetwork()
+);
+
+export default function useVotesRevealedEventsSummary() {
+  const { addError } = useContext(ErrorContext);
+
+  const { data, error, isFetching, refetch } = useQuery<
```
daywiss, line 20: Is there a reason why you destructure this and restructure it on the return, rather than just returning the useQuery directly?

src/hooks/useVotesRevealedEventsSummary.ts
```diff
@@ -0,0 +1,33 @@
...
+import {
+  queryVotesRevealedEvents,
+  VoteRevealed,
+} from "web3/get/queryVotesRevealedEvents";
+import { ErrorContext } from "common/context/ErrorContext";
+import provider from "common/utils/web3/createProvider";
+import createVoidSignerVotingContractInstance from "web3/createVoidSignerVotingContractInstance";
+import determineBlockchainNetwork from "web3/helpers/determineBlockchainNetwork";
+
+const contract = createVoidSignerVotingContractInstance(
```
daywiss, line 12: can u explain how this works and why not live in a hook? I guess im worried that there's like some uncontrolled re-instantiation of the web3 provider happening, when ideally the entire app should use the same instance.

src/features/vote/ActiveViewDetailsModal.tsx
```diff
@@ -38,24 +43,85 @@ interface Props {
   timestamp: string;
   ancData: string;
   unix: string;
+  committedVotes: VoteEvent[] | undefined | void;
+  revealedEvents: VoteRevealed[] | undefined | void;
```
Gamaranto, line 47: Why is there `void` here? I think its supposed to be used only expressions and is equal to `undefined` here

src/features/vote/ActiveViewDetailsModal.tsx
Gamaranto: If this is removing duplicates, you can do `const unique = [...new Set(array)];` as David pointed out

src/hooks/useVotesRevealedEventsSummary.ts
Gamaranto: Nothing to change, but one thing I find useful is to do `const {data, ...others} = useQuery(...)` and then returning them like `{someRelevantKey: data, ...others}`, which helps when using multiple `useQuery` hooks in the same component as you don't have a ton of `data` that you have to rename

Gamaranto (commented):
Would like to see the `useEffects` cleaned up, but otherwise LGTM

src/features/vote/ActiveViewDetailsModal.tsx
Tulun: Correct. We want to see how many people have voted -- we don't really care if they have resubmitted a vote.

src/features/vote/ActiveViewDetailsModal.tsx
Tulun: Yup, this could be abstracted out into a utility function. I'm not sure it would be used elsewhere, so it's more up to convention if we want the code close (IE: in this modal) or pulled out for testing.

src/hooks/useVotesRevealedEventsSummary.ts
Tulun: Yeah, David had suggested earlier we just return the default values then allow for the individual views to just handle naming convention as they see fit. It works either way.

And I often like returning what is currently relevant, as there is a ton of extra stuff useQuery actually returns from it's object. But we can just return the whole thing if we wish.

daywiss: its up to you, its just a style thing, just wanted to get your thoughts. i think its fine any of these ways

src/features/vote/ActiveViewDetailsModal.tsx
Tulun: You get a type error if you remove the void from useQuery. So it's one of the return types technically.

Screen Shot 2021-06-22 at 11 28 02 AM

src/hooks/useVotesRevealedEventsSummary.ts
Tulun: Basically, we create a read only version of the contract to read data from the contract. We want this because the user can see data from the contract not logged in.

src/features/vote/ActiveViewDetailsModal.tsx
Gamaranto: I see, I think that's because there is a `catch` in the queryFn, there is an `onError` callback that you can pass react-query to call `addError`, no need to change it though, I was just curious about the `void`

daywiss (approved):

PR #49, opened 2021-06-22
feat: Non 18 decs fix

Tulun:
Use non-18 decs for prices for old price identifiers.


package.json
```diff
@@ -23,7 +23,8 @@
     "@types/node": "^12.0.0",
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.0",
-    "@uma/core": "2.2.0-alpha.0",
+    "@uma/common": "^2.3.0",
+    "@uma/core": "^2.3.0",
```
nicholaspai, line 27: 2.4 is latest now!

nicholaspai (approved):
LGTM, try installing `uma/core: 2.4.0` though

PR #53, opened 2021-06-24
feat: Commit seed backup

Tulun:
Backup seed used for a particular round.

Apparently only the salt is required.
https://app.clubhouse.io/uma-project/story/1279/add-ability-to-view-voters-backup-commit-seed
Screen Shot 2021-06-23 at 3 00 07 PM

Added some padding to cells:

https://app.clubhouse.io/uma-project/story/1497/fix-spacing-between-rows

### DIFF
--- a/src/features/vote/ActiveViewDetailsModal.tsx
+++ b/src/features/vote/ActiveViewDetailsModal.tsx
@@ -80,6 +80,7 @@ const _ActiveViewDetailsModal: ForwardRefRenderFunction<
   const [revealPercentage, setRevealPercentage] = useState(NULL_PERCENTAGE);
   const [copySuccess, setCopySuccess] = useState(false);
   const [backupSeed, setBackupSeed] = useState("");
+
   // Note: because there is dynamic content, this will rebuild the tooltip for addressing the conditional
   // elements on the page. See ReactTooltip docs.
   useEffect(() => {
@@ -94,7 +95,6 @@ const _ActiveViewDetailsModal: ForwardRefRenderFunction<
       if (
         parsedJSON[votingAddress] &&
         parsedJSON[votingAddress][roundId] &&
-        parsedJSON[votingAddress][roundId] &&
         parsedJSON[votingAddress][roundId][uniqueIdentifier]
       ) {
         setBackupSeed(parsedJSON[votingAddress][roundId][uniqueIdentifier]);
@@ -102,7 +102,7 @@ const _ActiveViewDetailsModal: ForwardRefRenderFunction<
     } else {
       setBackupSeed("");
     }
-  }, [roundId, proposal, unix, votingAddress, ancData]);
+  }, [roundId, proposal, unix, votingAddress, ancData, committedVotes]);
 
   useEffect(() => {
     if (committedVotes && committedVotes.length && roundId && proposal) {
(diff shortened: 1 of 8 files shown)


Gamaranto (approved):
Minor comment, otherwise LGTM

src/features/vote/helpers/formatVoteDataToCommit.ts
```diff
@@ -22,12 +30,27 @@ export async function formatVoteDataToCommit(
...
+    // To prevent null errors in nested object, need to create these objects on the fly
+    // Based on if the user has voted before or not in a given round or using a given account
+    if (!newCommits[address]) {
+      newCommits[address] = {
+        [roundId]: {},
+      };
+    }
+    if (newCommits[address] && !newCommits[address][roundId]) {
+      newCommits[address][roundId] = {};
+    }
```
daywiss, line 46: this is really annoying, dealing with nested vals. theres a couple alternatives to consider. you can use lodash .has and .set to check for existence as well as set data deeply nested, will add to bundle size, but these can be installed as their own packages.

another option is to flatten the keys into a unique id and keep the object only 1 level deep.  You'd probably want a helper to generate the id based on the props you need to nest.  this is also nice since it plays well with localstorage.

just wanted to propose some things since it seems like most of the lines are dedicated to managing nested object

src/features/vote/helpers/formatVoteDataToCommit.ts
```diff
@@ -88,5 +113,7 @@ export async function formatVoteDataToCommit(
     })
   );
 
+  localStorage.setItem("backupCommits", JSON.stringify(newCommits));
```
daywiss, line 116: again not sure if its possible, but to return `newCommits` and do this in the caller

src/features/vote/helpers/formatVoteDataToCommit.ts
Tulun: Yep, nested objects suck. I'll look into if we can make this better.

src/features/vote/helpers/formatVoteDataToCommit.ts
Tulun: it could be done that way, although I'm not sure it would simplify it.

Maybe it would be better to just have a parallel function that is run and is atomic.

src/features/vote/ActiveViewDetailsModal.tsx
```diff
@@ -76,12 +79,31 @@ const _ActiveViewDetailsModal: ForwardRefRenderFunction<
...
   // Note: because there is dynamic content, this will rebuild the tooltip for addressing the conditional
   // elements on the page. See ReactTooltip docs.
   useEffect(() => {
     ReactTooltip.rebuild();
   });
 
+  useEffect(() => {
+    const commits = localStorage.getItem("backupCommits");
+    if (roundId && proposal && unix && votingAddress && commits && ancData) {
+      const parsedJSON = JSON.parse(commits);
```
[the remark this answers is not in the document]
Tulun: Makes sense.

src/features/vote/ActiveViewDetailsModal.tsx
```diff
@@ -76,12 +79,33 @@ const _ActiveViewDetailsModal: ForwardRefRenderFunction<
...
   });
 
+  useEffect(() => {
+    const commits = localStorage.getItem("backupCommits");
+    if (roundId && proposal && unix && votingAddress && commits && ancData) {
+      try {
+        const parsedJSON = JSON.parse(commits);
+        const uniqueIdentifier = `${proposal}~${unix}~${ancData}`;
+        if (
+          has(parsedJSON, `${votingAddress}.[${roundId}].${uniqueIdentifier}`)
```
daywiss, line 97: just fyi you can use arrays to represent your path ie: `[votingAddress,roundId,uniqueIdentifier]`, slightly easier than using a template string here. also if u want to keep the template string i dont think you need the `[]` around roundid

daywiss (approved):

PR #54, opened 2021-06-25
Feat: Better flow for committing if unsigned.

Tulun:
Modal does not open if key is unsigned. Error messages below pops up instead.

vote-error01

https://app.clubhouse.io/uma-project/story/1384/dapp-allows-user-to-go-through-the-voting-process-without-signing-make-the-dapp-present-an-error-message-when-the-user-clicks


daywiss (approved):
lgtm

PR #55, opened 2021-06-25
feat: Sign once a round

Tulun:
Implementation of this signing each round. Keeps it aligned with old dapp voting compatibility.

https://app.clubhouse.io/uma-project/story/863/never-or-almost-never-need-to-re-sign-when-using-the-same-browser-on-the-same-machine


src/common/helpers/currentSigningMessage.ts
```diff
@@ -0,0 +1,27 @@
...
+      message: SIGNING_MESSAGE_ONE,
+      hashedMessage: web3.utils.utf8ToHex(SIGNING_MESSAGE_ONE),
+    };
+  }
+
+  // Default return
+  return {
+    message: SIGNING_MESSAGE_ONE,
+    hashedMessage: web3.utils.utf8ToHex(SIGNING_MESSAGE_ONE),
+  };
```
daywiss, line 23: these branches are returning the exact same thing, doesnt seem right

src/App.tsx
```diff
@@ -36,35 +50,63 @@ function App(props: Props) {
...
-            const privateKey = derivePrivateKey(msg);
-            const publicKey = recoverPublicKey(privateKey);
-            key.privateKey = privateKey;
-            key.publicKey = publicKey;
-            key.roundMessage = message;
-
-            setSigningKeys((prevKeys) => {
-              return { ...prevKeys, [address]: key };
+      let keysInStorage =
+        localStorage.getItem(SIGNING_KEYS_STORAGE_KEY) || ({} as StorageKeys);
```
daywiss, line 60: its a bit wierd to mix string and json here, makes this variable a double type, why not use "{}" as the fallback so yo u can treat it as an empty object if undefined

src/common/helpers/getSigningKeys.ts
```diff
@@ -0,0 +1,52 @@
...
+//         ? signingKeys[hexMessage][votingAddress].privateKey
+//         : "";
+//       keys.publicKey = signingKeys[hexMessage][votingAddress]
+//         ? signingKeys[hexMessage][votingAddress].publicKey
+//         : "";
+//     }
+//   }
+
+//   return keys;
+// }
```
daywiss, line 33: you can prob clean this up

src/App.tsx
```diff
@@ -36,35 +50,63 @@ function App(props: Props) {
...
+
+      let keyExistsInStorage = false;
+      if (typeof keysInStorage === "string") {
+        keysInStorage = JSON.parse(keysInStorage) as StorageKeys;
+
+        keyExistsInStorage = has(keysInStorage, `${hexMessage}.${address}`);
+      }
+
+      if (keyExistsInStorage) {
```
daywiss, line 69: it looks like storage always overrides whats in memory, so local storage acts as source of truth.  as long as theres no way to get out of sync between memory and local storage i think thats ok.  is that a safe assumption?

src/App.tsx
```diff
@@ -36,35 +50,63 @@ function App(props: Props) {
...
+              localStorage.setItem(
+                SIGNING_KEYS_STORAGE_KEY,
+                JSON.stringify(keysToBackup)
+              );
+              setSigningKeys((prevKeys) => {
+                return { ...prevKeys, [address]: key };
+              });
+            })
+            .catch((err) => {
+              const error = new Error("Sign failed.");
```
daywiss, line 106: this obscures the actual error, what do u think about doing something like this `new Error("Sign failed: " + err.message);`

src/App.tsx
```diff
@@ -11,18 +11,32 @@ import { OnboardContext } from "common/context/OnboardContext";
...
+
 interface Props {
   queryClient: QueryClient;
 }
 
+export interface SigningKey {
+  publicKey: string;
+  privateKey: string;
+  roundMessage: string;
+  roundId: string;
```
mrice32, line 26: Two questions:
1. Why do we need these variables (roundId and roundMessage) here if they're effectively encoded into the mapping, itself?
2. Why do we need a concept of round hardcoded into the data structure? It seems like we could be agnostic to how the message is generated or on what frequency it changes and leave that up to the message generator, no?

src/App.tsx
```diff
@@ -36,35 +50,63 @@ function App(props: Props) {
...
+        keyExistsInStorage = has(keysInStorage, `${hexMessage}.${address}`);
+      }
+
+      if (keyExistsInStorage) {
+        const loggedInKey = keysInStorage[hexMessage][address];
+        setSigningKeys((prevKeys) => {
+          return { ...prevKeys, [address]: loggedInKey };
+        });
+      } else {
+        if (!keyExists || keyExists.roundMessage !== message) {
```
mrice32, line 75: Similar to the question above, since we're using the message to look the value up in the mapping -- why do we need to store it inside the data blob and then re-check it here?

src/App.tsx
```diff
@@ -36,35 +50,63 @@ function App(props: Props) {
...
+              key.roundMessage = message;
+              key.roundId = roundId;
+
+              let keysToBackup = {};
+              if (typeof keysInStorage === "string") {
+                const parsedJSON = JSON.parse(keysInStorage);
+                keysToBackup = { ...parsedJSON };
+              } else {
+                keysToBackup = { ...keysInStorage };
+              }
```
mrice32, line 94: What is this else case? Can localstorage return an object? When would that happen? Or is this just because we cast the `||` case above to `StorageKeys`?

src/common/helpers/getSigningKeys.ts
```diff
@@ -0,0 +1,52 @@
+import { SigningKeys } from "App";
+
+// export default function getSigningKeys(
```
mrice32, line 3: Commented code?

mrice32 (commented):
Looks good! A few minor comments

src/App.tsx
Tulun: Yeah that's fair.

src/common/helpers/currentSigningMessage.ts
Tulun: It is correct in this case.

If you can imagine we only change the message every N period, we add a hardcoded value. We could just add in the conditional statement later once a message changes, but it's to show how the code works after a message changes.

src/App.tsx
Tulun: The flow here is that you save new data into local storage and then update what's in state so react can re-render.

I think that's the only way this works. Because just referencing localStorage is insufficient for React to know to update state.

src/App.tsx
Tulun: Yep, we can do that. This was largely initially for when a user clicked cancel, but we can just return the error message.

src/App.tsx
Tulun: I believe this check exists solely if the user is on the app transitioning between rounds.

It's sort of an edge case to indicate they need to re-sign.

src/App.tsx
Tulun: It's a typescript thing. Basically unless I do a type check here, it won't let me spread over a string.

src/App.tsx
Tulun: I actually get a type error if I try to swap it over... rather annoyingly. TypeScript is sometimes really frustrating to work with

src/common/helpers/getSigningKeys.ts
Tulun: Removed!

src/common/helpers/getSigningKeys.ts
Tulun: Done!

src/App.tsx
daywiss: hmm ok i just tried this and it seemed to work:

```js
// parsing the json at the same time we fetch from storage
const keysInStorage = JSON.parse(localStorage.getItem(SIGNING_KEYS_STORAGE_KEY) || '{}') as StorageKeys;
// using array style path notation, saves some chars 
const keyExistsInStorage = has(keysInStorage, [hexMessage,address]);
```

src/App.tsx
daywiss: ok cool that makes sense

src/App.tsx
daywiss: it looks like this would mainly throw in the signer, but it could also throw in business logic for whatever reason.  we could also just log to console.error if you dont want to expose the user to the raw message.  just need a way to debug if someoen reports a problem imo

src/App.tsx
mrice32: It totally makes sense to store them, but it's unclear why they need to be associated with the signing key. Instead of doing checks on this data, isn't the more standard pattern to have a useMemo or useEffect that sets the computed message (or roundId) as a dependency?

Something like (oversimplified):
```ts
export interface SigningKey {
  publicKey: string;
  privateKey: string;
};

// Note: connectionStatus is some variable that changes when we the user changes something about their wallet connection.

useEffect(() => {
  // Compute message
  const message = getMessage(roundId);
  updateSigningKey(message);
}, [connectionStatus, roundId]);

```

The reason I bring this up is that I think the generation of the message and how we store it should ideally be completely abstracted away from one another. The storage should only care about mapping from message to keys, with no regard to how that message is generated. We only need to re-evaluate our signing keys when some input (either the signing address, connection status, or the message) changes.

Does that make sense? Is my understanding of the fundamental logic wrong?

mrice32 (commented):
Just one comment! https://github.com/UMAprotocol/voter-dapp/pull/55#discussion_r660781947

src/App.tsx
Tulun: It makes sense, but I am trying to think of how we'd organize this. Because it makes sense to keep both of these pieces of data close and in memory, as it's essentially part of the user's logged in state.

It helps because if the user is on the app when the round crosses over, we need to make sure we are doing some direct comparisons.

evaldofelipe:
we are able to merge it? @Tulun

Tulun:
> we are able to merge it? @Tulun

Not approved for this PR.

PR #64, opened 2021-09-02
fix: allow designated voting contract to be set more than once

daywiss:
### Motivation:

https://app.clubhouse.io/uma-project/story/2258/a-commit-transaction-was-successful-but-ui-shows-uncommitted

### Bug
What ended up happenign was on commit, it would not commit to the 2key contract. Deep in the code the cause was that the designated contract instance was undefined, but everything else looked correct. So there was an issue with instantiating the 2key contract instance. 

### Summary:
2 main changes here. In useVotingAddress there was a useEffect which looks like it could possibly override your voting address.
More likely the bug was in useVotingContract, where the designated voting contract could not be set more than once. Meaning if for some reason your voting address was undefined, it could never be set again once it became defined (at least thats my best guess). 

### Testing
Tested this in production with my account, as it would never send to the 2key voting contract, and now it does.

### DIFF
--- a/src/hooks/useVotingAddress.ts
+++ b/src/hooks/useVotingAddress.ts
@@ -31,8 +31,8 @@ export default function useVotingAddress(
         }).catch((err:Error)=>{
           console.error('Error getting designated voting address:',err)
         });
-    }else{
-      setVotingAddress(address);
+    } else {
+      setVotingAddress(null);
     }
   }, [address, signer, network]);
 
--- a/src/hooks/useVotingAddress.ts
+++ b/src/hooks/useVotingAddress.ts
@@ -28,9 +28,12 @@ export default function useVotingAddress(
             setVotingAddress(res);
             setHotAddress(address);
           }
+        }).catch((err:Error)=>{
+          console.error('Error getting designated voting address:',err)
         });
+    }else{
+      setVotingAddress(address);
     }
-    setVotingAddress(address);
   }, [address, signer, network]);
 
   return {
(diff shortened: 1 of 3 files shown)


src/hooks/useVotingContract.ts
```diff
@@ -29,13 +29,14 @@ export default function useVotingContract(
...
-        if (hotAddress && voterAddress) {
-          const dvc = createDesignatedVotingContractInstance(
-            signer,
-            voterAddress
-          );
-          setDesignatedVotingContract(dvc);
-        }
+      }
+      // we want this to be able to change when user switches accounts. previously this was only being able to be set once.
+      if (signer && hotAddress && voterAddress) {
```
daywiss, line 34: this was moved out of a nested if statement and into its own block. If you check the previous code you will see that if the outer if statement ran without running the inner statement, the inner statement would not have a chance to ever update again.

Tulun (approved):
LGTM!

src/hooks/useVotingAddress.ts
```diff
@@ -28,9 +28,12 @@ export default function useVotingAddress(
             setVotingAddress(res);
             setHotAddress(address);
           }
+        }).catch((err:Error)=>{
+          console.error('Error getting designated voting address:',err)
         });
+    }else{
```
mrice32, line 34: formatting?

mrice32 (approved):
LGTM! Very nice find!

PR #67, opened 2021-09-30
feat: Sign only once a round.

Tulun:
Motivation:

User should only need to sign once per round to reduce friction per account.

Details:

Created a useSigningKeys hook to hold the relevant state, and then have it instantiate a base state object based on whether or not the user has a signingKeys object in localStorage.

### DIFF
--- a/src/hooks/useSigningKeys.ts
+++ b/src/hooks/useSigningKeys.ts
@@ -19,7 +19,11 @@ export default function useSigningKeys(
   address: string | null,
   roundId: string
 ) {
-  const [signingKeys, setSigningKeys] = useState<SigningKeys>({});
+  const storageKeys = localStorage.getItem("signingKeys");
+
+  const [signingKeys, setSigningKeys] = useState<SigningKeys>(
+    storageKeys ? (JSON.parse(storageKeys) as SigningKeys) : {}
+  );
 
   const { addError } = useContext(ErrorContext);
 
--- a/src/hooks/useSigningKeys.ts
+++ b/src/hooks/useSigningKeys.ts
@@ -2,6 +2,10 @@ import { ethers } from "ethers";
 import { useState, useEffect, useContext } from "react";
 import { ErrorContext } from "common/context/ErrorContext";
 
+import { recoverPublicKey } from "features/vote/helpers/recoverPublicKey";
+import { derivePrivateKey } from "features/vote/helpers/derivePrivateKey";
+import usePrevious from "common/hooks/usePrevious";
+
 export interface SigningKeys {
   [key: string]: {
     publicKey: string;
@@ -10,13 +14,9 @@ export interface SigningKeys {
   };
 }
 
-import { recoverPublicKey } from "features/vote/helpers/recoverPublicKey";
-import { derivePrivateKey } from "features/vote/helpers/derivePrivateKey";
-import usePrevious from "common/hooks/usePrevious";
-
 export default function useSigningKeys(
   signer: ethers.Signer | null,
-  address: string,
+  address: string | null,
   roundId: string
 ) {
   const [signingKeys, setSigningKeys] = useState<SigningKeys>({});
@@ -52,7 +52,10 @@ export default function useSigningKeys(
             key.roundMessage = message;
 
             setSigningKeys((prevKeys) => {
-              return { ...prevKeys, [address]: key };
+              const updateKeys = { ...prevKeys, [address]: key };
+              localStorage.setItem("signingKeys", JSON.stringify(updateKeys));
+
+              return updateKeys;
             });
           })
           .catch((err) => {
@@ -71,5 +74,8 @@ export default function useSigningKeys(
     roundId,
   ]);
 
-  return signingKeys;
+  return {
+    signingKeys,
+    setSigningKeys,
+  };
 }
--- a/src/hooks/useSigningKeys.ts
+++ b/src/hooks/useSigningKeys.ts
@@ -6,8 +6,10 @@ import { recoverPublicKey } from "features/vote/helpers/recoverPublicKey";
 import { derivePrivateKey } from "features/vote/helpers/derivePrivateKey";
 import usePrevious from "common/hooks/usePrevious";
 
+// User has to sign once per round.
+// public and private signing keys is indexed by address, and we ask them to resign if the roundMessage changes.
 export interface SigningKeys {
-  [key: string]: {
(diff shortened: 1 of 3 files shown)


src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,87 @@
...
+            key.roundMessage = message;
+
+            setSigningKeys((prevKeys) => {
+              const updateKeys = { ...prevKeys, [address]: key };
+              localStorage.setItem("signingKeys", JSON.stringify(updateKeys));
+
+              return updateKeys;
+            });
+          })
+          .catch((err) => {
```
daywiss, line 67: might want to log this error to console

daywiss (approved):

src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,87 @@
...
+    roundMessage: string;
+  };
+}
+
+export default function useSigningKeys(
+  signer: ethers.Signer | null,
+  address: string | null,
+  roundId: string
+) {
+  const storageKeys = localStorage.getItem("signingKeys");
```
Gamaranto, line 24: reading from localstorage is possibly expensive, we could batch this call with the `useState` below and use the initialization function of `useState`, which would only run on mount, for ex
```
const [keys, setKeys] = useState(() => {
const storage = localStorage.getItem("keys");
// ... other stuff
})

**Review by Gamaranto — COMMENTED** [review]
just a small comment, otherwise lgtm

File: src/hooks/useSigningKeys.ts (hunk 2 above)
Comment (Tulun) [reply src/hooks/useSigningKeys.ts]: Sounds good, done!

File: src/hooks/useSigningKeys.ts [hunk 3]
```diff
@@ -0,0 +1,86 @@
+import { ethers } from "ethers";
```
mrice32, line 1: This seems a lot simpler, +1!

src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,86 @@
...
+  const previousAddress = usePrevious(address);
+
+  useEffect(() => {
+    if (
+      signer &&
+      address &&
+      roundId &&
+      (previousAddress === null || previousSigner === null)
+    ) {
+      const message = `UMA Protocol one time key for round: ${roundId}`;
```
mrice32, line 41: Is there any reason we can't just drop the round id from the message so people _never_ have to sign again? Or do we just want to slow roll this in case there are any issues?

src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const previousSigner = usePrevious(signer);
+  const previousAddress = usePrevious(address);
+
+  useEffect(() => {
+    if (
+      signer &&
+      address &&
+      roundId &&
+      (previousAddress === null || previousSigner === null)
```
mrice32, line 39: Can you explain what this does? Is it possible this could block updates if the user swaps accounts?

src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,86 @@
...
+    publicKey: string;
+    privateKey: string;
+    roundMessage: string;
+  };
+}
+
+export default function useSigningKeys(
+  signer: ethers.Signer | null,
+  address: string | null,
+  roundId: string
```
mrice32, line 22: In the calling component, it looks like we default this to `""`, which will technically trigger the if in the useEffect to be false. This works, but it's a little confusing (IMO) -- why not just allow it to be null like the others, not default it in the calling component, and assume that null means the data hasn't come in, just for the sake of consistency with the other data?

src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,86 @@
...
+            const error = new Error("Sign failed.");
+            addError(error);
+          });
+      }
+    }
+  }, [
+    signer,
+    address,
+    signingKeys,
+    addError,
```
mrice32, line 76: Is this a function? Are we sure this doesn't get compared by reference and causes the useEffect to run on every render because the function gets reinitialized on each run?

src/hooks/useSigningKeys.ts
```diff
@@ -0,0 +1,86 @@
...
+          .catch((err) => {
+            const error = new Error("Sign failed.");
+            addError(error);
+          });
+      }
+    }
+  }, [
+    signer,
+    address,
+    signingKeys,
```
mrice32, line 75: Similar to the comment about addError, can we verify that none of these are going to cause spurious calls to this? As far as I'm aware, non-primitives are compared by reference, which can cause this to get called on every render? A quick benchmark might be to put a console.log in this useEffect (outside of the if) and see if it's getting called constantly or only when account details change.

mrice32 (commented):
Looks great! Way simpler! A few comments on the implementation details.

src/hooks/useSigningKeys.ts
Tulun: Yeah I think it's largely due to being a bit conservative about the signing process.

Nothing technically. We'd just want to have a good system in place to allow the user to *force* a re-sign if they needed to, or invalidate it if we know its wrong.

src/hooks/useSigningKeys.ts
Tulun: Basically because of the way we use react-query, these values update in different render cycles. I believe this catch is in place to prevent the user having to sign twice.

src/hooks/useSigningKeys.ts
Tulun: Are you referring to roundId?

src/hooks/useSigningKeys.ts
Tulun: It shouldn't, as I believe addError is memoized.

src/hooks/useSigningKeys.ts
Tulun: I think you can already verify this by the number of times it asks you to sign, which is once :)

amateima (approved):

src/hooks/useSigningKeys.ts
mrice32: Would be nice to verify. It would be great, in general, like for us to have better reasoning around how often these useEffects run (and add comments about it!) and what types of changes trigger them. Otherwise, it's difficult to reason about the state changes that can cause the ifs to block or accept changes.

PR #68, opened 2021-10-12
feat: Sign message is static

Tulun:
Final update to sign message. This assume we change the message *in this file* and not in using an environment variable.


mrice32 (approved):

Tulun:
Potentially mergeable at our leisure @daywiss @mrice32

PR #73, opened 2021-11-15
fix: fix issue where it shows a vote was revealed when not

daywiss:
This pr fixes the reveal issue and refactors onboard init to ensure it and notifyjs only happens once on app load. this seems to generally improve performance of the app. 

The reveal bug: this would happen when you first load the page, you would see that you revealed when you hadnt yet.  I believe the main culprit here was sharing react-query caches between other similar calls to fetch revealed events. Some part of the app fetches all reveal events, to show statistics in the details for each vote. If this fills the cache, your vote would be "found" as revealed because all args on the reveal objects were checked *except* the address that voted.  There were some other minor issues cleaned up during this investigation.

refactoring onboard: this started as looking into performance issues with the app. I performed a similar change to the across app, moving onboard init out of react to ensure it only gets instanced once. This seems to have positive impact on the app as it feels slightly more responsive

### DIFF
--- a/src/common/context/OnboardContext.tsx
+++ b/src/common/context/OnboardContext.tsx
@@ -1,9 +1,9 @@
-import { createContext, useReducer, Dispatch, FC } from "react";
+import { createContext, useReducer, Dispatch, FC, useEffect } from "react";
 import { ethers } from "ethers";
-import { API as OnboardApi } from "bnc-onboard/dist/src/interfaces";
+import { API as OnboardApi, Wallet} from "bnc-onboard/dist/src/interfaces";
 import createOnboardInstance from "common/utils/web3/createOnboardInstance";
-import { Subscriptions } from "bnc-onboard/dist/src/interfaces";
-import { API as NotifyAPI } from "bnc-notify";
+import Notify, { API as NotifyAPI } from "bnc-notify";
+import Events from 'events'
 
 type Provider = ethers.providers.Web3Provider;
 type Address = string;
@@ -12,7 +12,7 @@ type Signer = ethers.Signer;
 
 type OnboardState = {
   provider: Provider | null;
-  onboard: OnboardApi | null;
+  onboard: OnboardApi;
   signer: Signer | null;
   network: Network | null;
   address: Address | null;
@@ -21,8 +21,23 @@ type OnboardState = {
   notify: NotifyAPI | null;
 };
 
+type ChainId = 1 | 42 | 1337;
+
+const getNetworkName = (chainId: ChainId) => {
+  switch (chainId) {
+    case 1: {
+      return "homestead";
+    }
+    case 42: {
+      return "kovan";
+    }
+    case 1337: {
+      return "test";
+    }
+  }
+};
+
 const SET_PROVIDER = "SET_PROVIDER";
-const SET_ONBOARD = "SET_ONBOARD";
 const SET_SIGNER = "SET_SIGNER";
 const SET_NETWORK = "SET_NETWORK";
 const SET_ADDRESS = "SET_ADDRESS";
@@ -33,7 +48,6 @@ const SET_NOTIFY = "SET_NOTIFY";
 
 export const actions = {
   SET_PROVIDER: SET_PROVIDER as typeof SET_PROVIDER,
-  SET_ONBOARD: SET_ONBOARD as typeof SET_ONBOARD,
   SET_SIGNER: SET_SIGNER as typeof SET_SIGNER,
   SET_NETWORK: SET_NETWORK as typeof SET_NETWORK,
   SET_ADDRESS: SET_ADDRESS as typeof SET_ADDRESS,
@@ -47,10 +61,6 @@ type Action =
   | {
       type: `${typeof SET_PROVIDER}`;
       payload: Provider | null;
-    }
-  | {
-      type: typeof SET_ONBOARD;
-      payload: OnboardApi | null;
     }
   | {
       type: typeof SET_SIGNER;
@@ -80,18 +90,27 @@ export type OnboardDispatch = Dispatch<Action>;
 type WithDelegatedProps = {
   [k: string]: unknown;
 };
+
+const events = new Events()
+const subscriptions = {
+  address: (addr: string | null) => {
+    events.emit('event','address',addr)
+  },
+  network: async (networkId: any) => {
+    events.emit('event','network',networkId)
+  },
+  wallet: async (wallet: Wallet) => {
+    events.emit('event','wallet',wallet)
+  },
+};
+const onboardInstance = createOnboardInstance(1 as any, subscriptions);
+
 function connectionReducer(state: OnboardState, action: Action) {
   switch (action.type) {
     case SET_PROVIDER: {
       return {
         ...state,
         provider: action.payload,
-      };
-    }
-    case SET_ONBOARD: {
-      return {
-        ...state,
-        onboard: action.payload,
       };
     }
     case SET_SIGNER: {
@@ -141,7 +160,7 @@ function connectionReducer(state: OnboardState, action: Action) {
 
 const INITIAL_STATE = {
   provider: null,
-  onboard: null,
+  onboard: onboardInstance,
   signer: null,
   network: null,
   address: null,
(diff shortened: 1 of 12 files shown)


src/common/context/OnboardContext.tsx
```diff
@@ -80,6 +90,21 @@ export type OnboardDispatch = Dispatch<Action>;
...
+    events.emit("event", "address", addr);
+  },
+  network: async (networkId: any) => {
+    events.emit("event", "network", networkId);
+  },
+  wallet: async (wallet: Wallet) => {
+    events.emit("event", "wallet", wallet);
+  },
+};
+const onboardInstance = createOnboardInstance(1 as any, subscriptions);
```
daywiss, line 106: instanciate single instance of onboard outside of a react component. subscribe only to events inside react component

src/common/context/OnboardContext.tsx
```diff
@@ -191,12 +201,94 @@ export const OnboardContext = createContext<TOnboardContext>(
...
+
 const OnboardProvider: FC<WithDelegatedProps> = ({
   children,
   ...delegated
 }) => {
   const [state, dispatch] = useReducer(connectionReducer, INITIAL_STATE);
 
+  useEffect(() => {
+    events.on("event", (event: string, data: any) => {
+      switch (event) {
```
daywiss, line 229: this code is almost identically copied from the previous subscription. some uneeded network calls are removed from the wallet event

src/common/hooks/useOnboard.ts
```diff
@@ -43,85 +24,6 @@ export default function useOnboard() {
...
-              payload: await ethersProvider.getNetwork(),
-            });
-          } else {
-            dispatch({ type: actions.SET_PROVIDER, payload: null });
-            dispatch({ type: actions.SET_NETWORK, payload: null });
-          }
-        },
-      };
-
-      connect(dispatch, network, subscriptions, onboard);
```
daywiss, line 111: because this is in a use effect with so many dependencies, i didnt trust that this was only being initialized once

src/common/web3/get/queryVotesRevealedEvents.ts
```diff
@@ -28,33 +29,28 @@ export const queryVotesRevealedEvents = async (
...
     address,
-    roundId,
+    roundId && ethers.BigNumber.from(roundId),
     identifier,
     time,
     price,
     null,
     numTokens
   );
+  const events = await contract.queryFilter(filter, VOTER_CONTRACT_BLOCK);
```
daywiss, line 39: just removed try catch. its caught by caller

src/common/web3/get/queryVotesRevealedEvents.ts
```diff
@@ -28,33 +29,28 @@ export const queryVotesRevealedEvents = async (
   // VoteRevealed(address,uint256,bytes32,uint256,int256,bytes,uint256)
   const filter = contract.filters.VoteRevealed(
     address,
-    roundId,
+    roundId && ethers.BigNumber.from(roundId),
```
daywiss, line 32: this was never being passed a round id, and was erroring unless converted to bn

src/features/vote/Vote.tsx
```diff
@@ -74,7 +74,7 @@ const Vote: FC<Props> = ({
   const {
     data: revealedVotes = [] as VoteRevealed[],
     refetch: refetchVoteRevealedEvents,
-  } = useVotesRevealedEvents(votingContract, votingAddress);
+  } = useVotesRevealedEventsRound(votingContract, votingAddress || hotAddress, roundId);
```
daywiss, line 77: two issues here. this was only working correctly with a 2 key, as votingAddress was soemtimes undefined when not using 2key. second the round id was not used, so you would be quering all revealed events, when we only cared about those for this round.

src/features/wallet/Wallet.tsx
```diff
@@ -242,7 +241,7 @@ const Wallet: FC<Props> = ({ signingKeys, refetchVoteSummaryData }) => {
               <Button
                 className="connect-btn"
                 onClick={() => {
-                  if (!initOnboard) setInitOnboard(true);
+                  connect().catch(console.error)
```
daywiss, line 244: with the onboard refactor, we dont need to re-init onboard, we can just call connect

src/hooks/useVotesRevealedEvents.ts
```diff
@@ -1,29 +1,17 @@
...
-    () => {
-      return queryVotesRevealedEvents(contract, address)
-        .then((res) => res)
-        .catch((err) => addError(err));
-    },
-    { enabled: contract !== null }
+  return useQuery<VoteRevealed[] | undefined | void>(
+    ["votesRevealedEvents", address],
+    () => queryVotesRevealedEvents(contract, address),
+    { enabled: !!contract && !!address }
```
daywiss, line 15: this was just refactored to be shorter and remove the error notification

src/hooks/useVotesRevealedEventsRound.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default function useVotesRevealedEvents(
+  contract: ethers.Contract | null,
+  address: string | null,
+  roundId?: string
+) {
+  return useQuery<VoteRevealed[] | undefined | void>(
+    ["votesRevealedEventsByRound", address, roundId],
+    () => queryVotesRevealedEvents(contract, address, roundId),
+    { enabled: !!contract && !!roundId && !!roundId.length && !!address }
```
daywiss, line 16: this was added as a way to query with a round id parameter

Tulun (approved):
Awesome job.

src/features/vote/RevealPhase.tsx
```diff
@@ -204,7 +204,7 @@ const RevealPhase: FC<Props> = ({
                                 }
                               })
                               .catch((err) => {
-                                console.log("err in snapshot", err);
+                                console.error("err in snapshot", err);
```
mrice32, line 207: Can we add err.stack to the printout? I've found it's very hard to locate the line where the error is happening without this. If there's stuff in the error you want to keep that's not in the stack, maybe we could print the stack separately?

src/common/web3/get/queryCurrentRoundId.ts
```diff
@@ -7,6 +7,6 @@ export const queryCurrentRoundId = async (contract: ethers.Contract) => {
       return roundIdToString;
     }
   } catch (err) {
-    console.log("err", err);
+    console.error("err", err);
```
mrice32, line 10: Same here -- would be good to have err.stack.

src/features/vote/Vote.tsx
mrice32: Wow, this is a big fix. +1

src/hooks/useVotesRevealedEvents.ts
mrice32: Does contract also need to be part of the key?

mrice32 (approved):
LGTM! only minor comments

src/features/vote/RevealPhase.tsx
daywiss: logging the entire error object will print the stack out without explicitly needing to print err.stack

src/hooks/useVotesRevealedEvents.ts
daywiss: I would avoid using contract in the key since its not a primitive value and theres a chance the contract instance could change without changing the underlying contract address or user address, triggering an unecessary refetch.   I would need to be really sure that `contract` is not changing due to some weird useEffect where its initialized.

src/common/context/OnboardContext.tsx
```diff
@@ -21,8 +21,23 @@ type OnboardState = {
   notify: NotifyAPI | null;
 };
 
+type ChainId = 1 | 42 | 1337;
+
+const getNetworkName = (chainId: ChainId) => {
+  switch (chainId) {
+    case 1: {
+      return "homestead";
```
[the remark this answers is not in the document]
daywiss: im not sure where this comes from, it was copied and pasted from another part of the code. i will look into this and see if its right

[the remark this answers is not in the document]
daywiss: this looks like it comes from the [ethers provider docs](https://docs.ethers.io/v5/api/providers/api-providers/#InfuraProvider) `provider = new InfuraProvider("homestead", projectId);`.  I've searched around and the `network.name` isnt used anywhere, so i think its ok to just leave this.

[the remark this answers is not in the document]
mrice32: Gotcha, just seems a little odd.

src/features/vote/RevealPhase.tsx
mrice32: My understanding (that could be wrong) is that these are different. I've seen this in node.

Stringifying an error gives you the message + "name": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString.

Printing out the stack gives you the line where the error was thrown: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack.

mrice32 (commented):
Replied to the error.stack thread above.

PR #75, opened 2021-12-16
fix: App doesn't crash when on wrong network.

Tulun:
Passes in empty string instead.

Prompt is done in MetaMask -- we may want a more in app solution, but that's probably a different ticket.


daywiss (approved):
awesome, seems to work

amateima (approved):

PR #76, opened 2021-12-16
fix: check if rewards events changes.

Tulun:
Recheck balance if rewards events changes.


daywiss:
i think this doesnt fix the right thing, we might need to talk to tom to see if anyone can reproduce the issue

Tulun:
> i think this doesnt fix the right thing, we might need to talk to tom to see if anyone can reproduce the issue

That's fine, I can close it. But yeah I can't seem to repro the issue, so I'm going to move the ticket to unscheduled.

PR #77, opened 2022-01-02
fix: fix committing when anc data fails to decode

daywiss:

### DIFF
--- a/src/features/vote/helpers/formatVoteDataToCommit.ts
+++ b/src/features/vote/helpers/formatVoteDataToCommit.ts
@@ -47,16 +47,7 @@ export async function formatVoteDataToCommit(
         datum.identifier = el.idenHex;
         // datum.time = el.time;
         datum.time = el.timeBN.toNumber();
-        let ancData = "";
-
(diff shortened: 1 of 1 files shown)


src/features/vote/helpers/formatVoteDataToCommit.ts
```diff
@@ -47,16 +47,7 @@ export async function formatVoteDataToCommit(
...
         // datum.time = el.time;
         datum.time = el.timeBN.toNumber();
-        let ancData = "";
-
-        // anc data is set to - or N/A in UI if empty, convert back to 0x.
-        if (el.ancillaryData === "-" || el.ancillaryData === "N/A") {
-          ancData = "0x";
-        } else {
-          ancData = web3.utils.utf8ToHex(el.ancillaryData);
-        }
```
Tulun, line 57: Nice, so we can't assume this conversion works. I'd suggest we wrap this call in a try / catch and set it to your solution -- so we know we aren't breaking anything else.

Tulun (approved):

PR #78, opened 2022-01-03
fix: don't convert anc data. Reference ancHex

Tulun:
Fixes reveal issue with anc data that's not UTF-8.


daywiss (approved):
Thanks

PR #83, opened 2022-03-01
fix: Added margins

Tulun:
Should fix this: https://app.shortcut.com/uma-project/story/4439/adjust-the-columns-in-the-voter-dapp-so-words-are-not-so-close-to-each-column

Wait till a vote for testing.


daywiss (commented):
lgtm i guess we can just wait for vote to approve until we can fix testing

PR #84, opened 2022-03-03
fix: change scroll to auto

daywiss:


Gamaranto (approved):

Tulun (approved):

chrismaree (approved):
looks good in all places except when I click the gear on my account:

image


I was not able to test the process of committing a vote. we'll need to check that part during a voting cycle.

PR #86, opened 2022-03-24
feat: replace header and footer with website versions

amateima:
Signed-off-by: amateima <amatei@umaproject.org>

Replace the header and the footer with the components used on the website and on UMAverse


daywiss (commented):
style looks slightly different between sites
voter dapp:
image

umaverse
image

otherwise looks great to me

daywiss:
im also seeing some wierd sidebar in mobile view
image

also the footer goes on really long:
image

i dont see these issues in the prod site

amateima:
@daywiss 

🟢 footer issue solved

I tried to fix quickly the issue with the white space on the right side of the website, but I need to spent more time on it. We should make a separate issue. The `downship` dropdown component is the one that causes the problem. It doesn't play well with the overflow scrolling behaviour of the table

Screenshot 2022-03-25 at 14 30 11

daywiss (approved):

Tulun (approved):
Code wise seems totally fine.

Slightly out of alignment with the way the app is coded as you just ported over some of the stuff in some ways but I think it's fine.

PR #88, opened 2022-03-29
feat: Change text if rewards claimable but not claimed in past request

Tulun:
Screen Shot 2022-03-29 at 12 30 05 PM

### DIFF
--- a/src/features/vote/PastRequests.tsx
+++ b/src/features/vote/PastRequests.tsx
@@ -65,6 +65,8 @@ const PastRequests: FC<Props> = ({
     unix: "",
   });
 
+  console.log("pastRequests", pastRequests);
+
   useEffect(() => {
     // Handle past requests differently depending on if user is logged in or not.
     if (voteSummaryData.length && roundId) {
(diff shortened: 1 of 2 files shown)


src/features/vote/PastRequests.tsx
```diff
@@ -65,6 +65,8 @@ const PastRequests: FC<Props> = ({
     unix: "",
   });
 
+  console.log("pastRequests", pastRequests);
```
daywiss, line 68: can remove

src/features/vote/PastRequests.tsx
```diff
@@ -136,7 +138,11 @@ const PastRequests: FC<Props> = ({
                     <div>{el.vote}</div>
                   </td>
                   <td>
-                    <div>{el.reward}</div>
+                    <div>
+                      {el.vote === el.correct && el.reward === "N/A"
+                        ? "Rewards Claimable"
```
daywiss, line 143: why not change `el.reward` to say "Rewards Claimable" rather than "N/A"?  i think if thats the case you dont have to edit the view, though im not sure

daywiss (approved):

src/features/vote/PastRequests.tsx
Tulun: Lightest change as a lot of other stuff is looking for "-" or "N/A"

PR #89, opened 2022-04-04
fix: limit width of vote TD.

Tulun:
Screen Shot 2022-04-04 at 1 48 41 PM


daywiss (approved):

Tommy1231232 (changes requested):
@Tulun looks great but can we round the "Earned rewards" column to two decimal places?

Tommy1231232:
> @Tulun looks great but can we round the "Earned rewards" column to two decimal places?

I know this PR isn't for that but if you can do it here that'd be great

PR #95, opened 2022-06-13
feat(voter-dapp): add walletconnect

ryanwolhuter:
**Motivation**

We would like to support the Gnosis Safe multisig in the dApp, and we are able to do so via WalletConnect. Our WalletConnect provider Onboard provides this functionality, and we can enable it with some configuration.

**Solution**

* Add Onboard configuration for WalletConnect


src/common/config.ts
```diff
@@ -35,16 +35,16 @@ export const OLD_VOTING_CONTRACT_ADDRESSES = [
...
+    network ? network?.name : "mainnet"
+  }.infura.io/v3/${infuraId}`;
 
   return {
     onboardConfig: {
       apiKey: process.env.REACT_APP_PUBLIC_ONBOARD_API_KEY || "",
       onboardWalletSelect: {
         wallets: [
-          { walletName: "metamask", preferred: true },
+          { walletName: "metamask" },
```
ryanwolhuter, line 47: I have removed `preferred: true` because this option hides the other wallet options behind a "show more" menu instead of making them readily visible.

ryanwolhuter:
I have tested this implementation and unfortunately I can confirm that it does not work.

Onboard works correctly for connecting to the dapp via a WalletConnect enabled wallet (able to view balance, rewards, etc.). However, the signing logic is particular to how MetaMask behaves through Onboard, and therefore the signing fails when connecting with WalletConnect.

It's worth noting that we are currently using an old (now deprecated) version of Onboard, and they have since completely overhauled their logic and API in the latest version. It's possible that the new version of Onboard does a better job at generalizing the signer interface. Upgrading Onboard would involve significant changes to the dapp though, as they have made several breaking changes to their API.

chrismaree (approved):

PR #96, opened 2022-06-21
feat(voter-dapp): add title and description to non-UMIPs price requests

ryanwolhuter:
**Motivation**

Currently, we only have logic for showing a title and a description in active and past requests for UMIPs. We would like to have titles and descriptions for non-UMIP requests as well, and this data can be derived from existing sources:

* For approved price identifiers, we can look up the summary and link to the relevant UMIP in a table of the data found in [our docs](https://docs.umaproject.org/resources/approved-price-identifiers).
* For Polymarket requests, we can decode the ancillary data and use the title and description that Polymarket requests provide in there.

**Summary**

* Add a helper function that determines what the description and title should be from the above constraints and data.
* Use this helper function in the past and active request view details modals.


src/common/apollo/queries.ts
```diff
@@ -37,6 +37,7 @@ export const PRICE_REQUEST_VOTING_DATA = gql`
       orderBy: $orderBy
       orderDirection: $orderDirection
     ) {
+      ancillaryData
```
ryanwolhuter, line 40: Currently we don't fetch the ancillary data in this request to the graph. I've added the field here because we need the ancillary data to get the title and description for Polymarket requests.

src/common/helpers/formatPriceRequestVoteData.ts
```diff
@@ -22,6 +22,7 @@ export interface FormattedPriceRequestRounds {
     uniqueClaimersPctOfReveals: string;
     time: number;
     voterRewards: string;
+    ancillaryData: string;
```
ryanwolhuter, line 25: We need to pass along the ancillary data when formatting so that it can be used later.

src/common/helpers/proposalTitleAndDescription/ancillaryData.ts
```diff
@@ -0,0 +1,48 @@
+import { ethers } from "ethers";
+
+export function decodeAncillaryDataHexString(ancillaryDataHexString: string) {
```
ryanwolhuter, line 3: I've extracted this logic to a helper function which we can use elsewhere going forward.

src/common/helpers/proposalTitleAndDescription/ancillaryData.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!ancillaryDataHexString.startsWith("0x")) {
+    return null;
+  }
+  return ethers.utils.toUtf8String(ancillaryDataHexString);
+}
+
+export function getTitleFromAncillaryData(
+  decodedAncillaryData: string,
+  titleIdentifier = "title:",
```
ryanwolhuter, line 12: Currently Polymarket uses "title:" and "description:" to identify these fields. I've made these default params in case we need to be flexible in the future.

package.json
```diff
@@ -61,7 +61,8 @@
     "test": "react-scripts test",
     "test:coverage": "yarn test --coverage --watchAll=false",
     "eject": "react-scripts eject",
-    "reveal": "ts-node ./manualreveal"
+    "reveal": "ts-node ./manualreveal",
+    "parse-approved-identifiers": "ts-node ./src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTableParser.ts"
```
ryanwolhuter, line 65: This script parses the approved identifiers in the approved identifiers table text file https://raw.githubusercontent.com/UMAprotocol/docs/master/docs/uma-tokenholders/approved-price-identifiers.md

We will automate this process for UMA 2.0

src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTableParser.ts
```diff
@@ -0,0 +1,30 @@
+import fs from "fs";
```
ryanwolhuter, line 1: A simple script for parsing the approved identifiers data found in https://raw.githubusercontent.com/UMAprotocol/docs/master/docs/uma-tokenholders/approved-price-identifiers.md

We will automate this process for UMA 2.0

chrismaree (commented):
One bit of feedback from playing with the UI. We have this UI element for links to UMIPS

image


(see purple arrow) that points to the UMIP for the given request. it would be good to use this in your implementation here:
image

src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTable.json
```diff
@@ -0,0 +1,1222 @@
+[
```
chrismaree, line 1: This is awesome and is. a great temp solution until we've improved the design of how we do voting issues.  however, from playing with the UI iI realized that the list you tok from is. a bit out of date. in particular, you are missing `ACROSS-V2`

image

see the umip here: https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-157.md

src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTable.txt
```diff
@@ -0,0 +1,244 @@
+|AAVEUSD| The price of AAVE in USD.| [UMIP-57](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-57.md)
```
chrismaree, line 1: why do you need both this and the json file?

package.json
chrismaree: this is awesome! great way to do this. can you add something to the docs around how to run this?

src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTable.json
chrismaree: actually the right solution is to update the docs then re-run the script, right?

ryanwolhuter:
> One bit of feedback from playing with the UI. We have this UI element for links to UMIPS
> 
> image
> 
> (see purple arrow) that points to the UMIP for the given request. it would be good to use this in your implementation here: image

Thanks for this Chris, I've updated the implementation as you suggested.

src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTable.json
ryanwolhuter: Yeah that's right, I'm just working with what's defined in the docs since we need to keep those in sync. I've set the script up to download that file with one command, so for now we can just run that script whenever the docs are updated. Will be easy to automate this in the future as well.

src/common/helpers/proposalTitleAndDescription/approvedIdentifiersTable.txt
ryanwolhuter: The file downloaded from GitHub is raw markdown and needs to be parsed to JSON before we can use it. I'm leaving it here as a paper trail, but we could just as easily add a line at the end of the script to delete the file when we're done generating the JSON file.

package.json
ryanwolhuter: Absolutely will do.

src/common/helpers/requestMetaData/approvedIdentifiersTableParser.ts
```diff
@@ -0,0 +1,91 @@
...
+    umipLink.indexOf("(") + 1,
+    umipLink.indexOf(")")
+  );
+
+  return { number, url };
+}
+
+async function downloadFile(url: string, targetFile: fs.PathLike) {
+  return await new Promise((resolve, reject) => {
+    https
```
daywiss, line 67: interesting, never used this directly before. nice way to not need extra deps

daywiss (approved):
looks great, just wondering though how do you see this script to generate identifier tables being run? do we need some process (manual or automated ) to keep this up to date?

src/common/helpers/requestMetaData/approvedIdentifiersTableParser.ts
ryanwolhuter: Glad you like it!

ryanwolhuter:
> looks great, just wondering though how do you see this script to generate identifier tables being run? do we need some process (manual or automated ) to keep this up to date?

Thanks man! 

So right now I'm thinking that this will be done manually, since we already have a manual process for updating the table of approved identifiers in the docs. So right now, I imagine that when the docs get updated, they can let a dev know and they can run the script to re-generate the file and then commit the new file. 

Of course I would prefer for this to be automated, but at the moment that will be complicated because it would involve listening for changes in the docs repo, and then making and committing changes in this repo (not a fun thing to automate).

Chris and I figured that the manual process is appropriate for the time being, since we're planning on totally overhauling how the price identifiers will be registered and queried for UMA 2.0.

Keen to hear what you think.

src/common/helpers/requestMetaData/approvedIdentifiersTableParser.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  await downloadFile(url, markdownPath);
+
+  // read and parse the markdown file from GitHub
+  const contents = fs.readFileSync(markdownPath, "utf-8");
+  const lines = contents
+    .split("\n")
+    // ignore the first 8 lines (headmatter and table headers)
+    .slice(8);
+  const parsedLines = JSON.stringify(lines.map(parseLine));
```
mrice32, line 22: nit: you can make this output formatted json by adding these args:

```suggestion
  const parsedLines = JSON.stringify(lines.map(parseLine), undefined, 2);
```

src/common/helpers/requestMetaData/approvedIdentifiersTableParser.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  for (let i = 0; i < line.length; i++) {
+    if (line[i] === "|") {
+      dividerIndices.push(i);
+    }
+  }
+
+  const title = line.substring(dividerIndices[0] + 1, dividerIndices[1]);
+  const summary = line.substring(dividerIndices[1] + 2, dividerIndices[2]);
+  const markdownUmipLink = line.substring(dividerIndices[2] + 2, line.length);
```
mrice32, line 42: tiny nit (you don't need to change this if you prefer the way it's done now): it seems like this block might be a bit simpler by doing something like:

```ts
const [titleSection, summarySection, markdownUmipLinkSection] = line.split("|");
const title = titleSection.slice(1);
const summary = summarySection.slice(2);
const umipLink = (parseMarkdownUmipLink(markdownUmipLinkSection.slice(2));
```

src/common/helpers/requestMetaData/getRequestMetaData.ts
```diff
@@ -0,0 +1,71 @@
...
+    return {
+      title: identifierDetails?.title ?? proposal,
+      description:
+        identifierDetails?.summary ??
+        "No description was found for this request.",
+      umipUrl: identifierDetails?.umipLink.url,
+    };
+  }
+
+  // if the previous checks fail, we assume we are dealing with a Polymarket request
```
mrice32, line 55: Wouldn't polymarket requests match the `YES_OR_NO_QUERY` identifier and get captured in the `isApprovedIdentifier` check above?

mrice32 (commented):
Looks awesome! This is really great code. Minor nits and comments

src/common/helpers/requestMetaData/ancillaryData.ts
```diff
@@ -0,0 +1,48 @@
+import { ethers } from "ethers";
+
+export function decodeAncillaryDataHexString(ancillaryDataHexString: string) {
+  if (!ancillaryDataHexString.startsWith("0x")) {
+    return null;
```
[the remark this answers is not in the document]
ryanwolhuter: I actually implemented it with an error the first time round. Looking back at it now I definitely should have stuck with my first instinct. I'll update this with more robust error catching, and also throw a different error for incorrect usage.

src/common/helpers/requestMetaData/approvedIdentifiersTableParser.ts
ryanwolhuter: This is awesome! Would definitely have done this if I had known about it. Will update 👍

src/common/helpers/requestMetaData/getRequestMetaData.ts
ryanwolhuter: You're absolutely right. I missed this because I didn't know that YES_OR_NO_QUERY is both. 

I've created an array of identifiers to ignore and treat as Polymarket requests instead. We can expand it as needed with other items like this.

Do you know of others like this? Or could you point me in the direction of where to find them?

mrice32: I'm a little confused with how you're describing it -- what do you mean when you say that `YES_OR_NO_QUERY` is both?

Below is probably more information than you need! But I just wanted to detail the landscape of UMIPs, governance requests, and ancillary data in case any of that was unclear. Feel free to ignore if you already understand all of this 🙂.

The way I think about it is:
- Everything that gets requested to the DVM is either a pre-registered identifier or a governance request.
- Most requests will have _some_ ancillary data. That ancillary data _should_ be formatted according to this document: https://docs.google.com/document/d/1zhKKjgY1BupBGPPrY_WOJvui0B6DMcd-xDR8-9-SPDw/edit#heading=h.220c6fyb8tgg.
- The fields that are _required_ (if any) for a particular identifier in the ancillary data should be specified in the UMIP for the identifier (most do not require ancillary data, but this is becoming more common). Example for [YES_OR_NO_QUERY](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-107.md).
- My guess is that we're only interested in changing how we display the request for `YES_OR_NO_QUERY`, since this is the _only_ one that polymarket uses, and it's the only one with an obvious title, etc. Others have ancillary data that is really just extra info for voters, but don't have display-related info.

src/common/helpers/requestMetaData/getRequestMetaData.ts
```diff
@@ -0,0 +1,79 @@
...
+      umipUrl: umip?.umipLink,
+    };
+  }
+
+  // if we are dealing with a request for an approved price identifier, get the title, description and UMIP url from the hard-coded approvedIdentifiersTable json file
+  // we know we are dealing with a request for an approved price identifier if the proposal matches an approved identifier's title
+  const identifierDetails = approvedIdentifiers.find(
+    (id) => proposal === id.title
+  );
+  const identifiersToIgnore = [
```
mrice32, line 44: This is totally a minor nit, but ignore seems like an odd term. it's almost like the opposite: we're selecting it to extract info from the ancillary data. Ignoring sounds like we're going to exclude it from the UI or something. I have no problems with the actual code here, it just conceptually feels a little off. No changes necessarily required if you're happy with this language, just wanted to note that it was a bit confusing to me as a reader.

mrice32 (approved):
LGTM!

src/common/helpers/requestMetaData/getRequestMetaData.ts
ryanwolhuter: Thanks very much for the explanation! I was trying to ask if there are any other special cases, i.e. approved identifiers that Polymarket uses which we should rather treat as Polymarket requests. 

Since `YES_OR_NO_QUERY` is the only one, and I do agree that the language I used was confusing, I've changed the implementation to simply treat `YES_OR_NO_QUERY` as a Polymarket request instead. I've removed the confusingly named array, and added a comment explaining the situation.

PR #97, opened 2022-07-22
fix: bump past request number, reduce block range for event queries

daywiss:

## Motivation
Event errors were showing up in console requesting too many events. Also past requests would not ever load during this vote with 5 active votes.

```
err Error: processing response error (body="{\"jsonrpc\":\"2.0\",\"id\":52,\"error\":{\"code\":-32005,\"message\":\"query returned 
more than 10000 results\"}}", error={"code":-32005}, requestBody="{\"method\":\"eth_getLogs\",\"params\":
[{\"fromBlock\":\"0xb539e7\",\"toBlock\":\"latest\",\"address\":\"0x8b1631ab830d11531ae83725fda4d86012eccd77\",
```

## Solution
Code was copied from the sdk which retries event queries and adjust block range dynamically to prevent event error. Also doubled the past queried events because it was set to 5, and 5 active votes are going on, so our query only included active requests, leaving the pending array empty.

### DIFF
--- a/src/common/components/navbar/Navbar.tsx
+++ b/src/common/components/navbar/Navbar.tsx
@@ -126,7 +126,6 @@ const HEADER_LINKS: IHeaderLink[] = [
   {
     key: "Vote",
     component: ({ path }) => {
-      console.log({ path });
       return (
         <UI.NavLink to="/" active={path === "/"}>
           Vote
--- a/src/common/web3/get/queryVotesCommittedEvents.ts
+++ b/src/common/web3/get/queryVotesCommittedEvents.ts
@@ -1,6 +1,7 @@
 import { ethers } from "ethers";
 import { VOTER_CONTRACT_BLOCK } from "common/config";
 import assert from "assert";
+import * as utils from "common/utils/events"
 
 import { VoteEvent } from "../types.web3";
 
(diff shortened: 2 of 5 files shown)


src/common/components/navbar/Navbar.tsx
```diff
@@ -126,7 +126,6 @@ const HEADER_LINKS: IHeaderLink[] = [
   {
     key: "Vote",
     component: ({ path }) => {
-      console.log({ path });
```
mrice32, line 129: nice

src/common/web3/get/queryVotesCommittedEvents.ts
```diff
@@ -35,25 +36,34 @@ export const queryVotesCommittedEvents = async (
...
-    });
-  } catch (err) {
-    console.log("err", err);
+  let events = []
+  let rangeState = utils.rangeStart({
+    startBlock:VOTER_CONTRACT_BLOCK,
+    endBlock: await contract.provider.getBlockNumber(),
+  })
+  while(!rangeState.done){
+    try{
```
mrice32, line 45: formatting here and elsewhere

mrice32 (approved):
Nice and simple! Didn't review the copied code, but everything else looks great. Thanks for adding this!

src/common/components/navbar/Navbar.tsx
ryanwolhuter: +1

ryanwolhuter:
Nice work!

ryanwolhuter (approved):
I agree with Mark. Also didn't review the copied code, but your implementation makes sense to me.

